PR #132, opened 2013-08-15
Cmake isnan isinf checks

JoakimSoderberg:
Changed strategy for looking for isnan/isinf that doesnt tie it to uclibc like in #131.


brian-armstrong:
This change appears to be necessary on Android as well for what it's worth, though it might not be after API 21.

cmake/jansson_config.h.cmake
```diff
@@ -57,6 +57,9 @@
 /* If locale.h and localeconv() are available, define to 1, otherwise to 0. */
 #define JSON_HAVE_LOCALECONV @JSON_HAVE_LOCALECONV@
 
-
```
AllenX2018: which case are "isnan" and "isinf" not available in?

Issue #133, opened 2013-08-16
Understanding problem for newcomers


igromanru:
Hello, I got a problem to understand how it´s works. Can you please extend the documentation on some more examples? Something simple like parse an object to json, write it to a file and read it back to an object.

RoyiAvital:
I'd be happy to see how to encode structs (Array of structs) into file.

PR #214, opened 2014-12-26
adding simple-parse example

rdpoor:
In partial response to #133, I've created `examples/simple_parse.c`, a simple standalone program that parses a line of text into a json_t structure and prints it.  (As a by-product, simple_parse.c has a `print_json()` method that recursively walks and prints a `json_t` object.)

What's missing is a makefile for the example directory and extensions to the other makefiles -- I'm not familiar enough with CMake or Makefile.am to confidently create them.


rdpoor:
Pleased to help.  jansson is a great tool.

Issue #153, opened 2014-01-08
[feature request] Function printing the error location in source json would be beneficial


yurivict:
Here is an example what it might print:

<pre>
                                 {["a","b","c",]}
 (expecting array element) --------------------^
</pre>

This would be very convenient for quick identification of problems in jsons.

JoakimSoderberg:
I added a proof of concept. The errors are not as specific as described in this ticket, but that's inherent from how Jansson parses I guess. Anyway, it simply uses the contents of `json_error_t`

JoakimSoderberg:
Oh and btw, does `error.position` refer to the bytes or character points for UTF-8?

akheron:
`error.position` is just plain bytes.

PR #218, opened 2015-01-14
Friendly error

JoakimSoderberg:
This is a proof of concept for printing a helper "error indicator" discussed in #153 

I made it so that you get the buffer back in this implementation, but a simple print function could be done as well. And I'm not sure if we should use malloc or not?

I added a use in the `simple_parser` test program for an easy way to test it with different input.

Example output:
```
Type some JSON > {["a","b","c",]}
Error:
{["a","b","c",]
  ^~~~~~ (string or '}' expected near '[')
```


akheron:
In my opinion using `malloc` is OK. Another option would be to use a user-supplied buffer, but it's somewhat harder to use.

I'll add diff comments next.

examples/simple_parse.c
```diff
@@ -158,7 +159,10 @@ json_t *load_json(const char *text) {
     if (root) {
         return root;
     } else {
-        fprintf(stderr, "json error on line %d: %s\n", error.line, error.text);
+		char *detailed = json_error_get_detailed(&error, text, 1);
+        //fprintf(stderr, "json error on line %d: %s\n", error.line, error.text);
+		fprintf(stderr, "Error:\n%s\n", detailed);
+		free(detailed);
```
akheron: Tabs vs spaces discrepancy here.

src/error.c
```diff
@@ -61,3 +61,135 @@ void jsonp_error_vset(json_error_t *error, int line, int column,
...
+
+    if (snprintf(s, len, "%*s", (int)len, start) < 0) {
+        free(s);
+        return NULL;
+    }
+
+    return s;
+}
+
+char *json_error_get_arrow(json_error_t *error, const char *src, int length, int color)
```
akheron: `int color` could be `size_t flags` instead for forwards compatibility.

akheron: And then the arrow length could also be encoded to flags...

src/error.c
```diff
@@ -61,3 +61,135 @@ void jsonp_error_vset(json_error_t *error, int line, int column,
...
+                error->column, "",
+                length, length, padchars)) < 0) {
+        goto fail;
+    }
+
+    offset += ret;
+
+    #ifndef _WIN32
+    if (color) {
+        if (snprintf(&msg[offset], msglen - offset, "%s", "\x1b[0m\x1b[0m") < 0) {
```
akheron: Using colored output is a cool idea, I like it :)

src/error.c
```diff
@@ -61,3 +61,135 @@ void jsonp_error_vset(json_error_t *error, int line, int column,
...
+    }
+    #endif // _WIN32
+
+    return msg;
+fail:
+    if (msg) free(msg);
+    return NULL;
+}
+
+char *json_error_get_detailed(json_error_t *error, const char *src, int color)
```
akheron: `color` could also be `flags` here.

src/jansson.h
```diff
@@ -126,6 +126,9 @@ typedef struct {
     char text[JSON_ERROR_TEXT_LENGTH];
 } json_error_t;
 
+char *json_error_get_detailed(json_error_t *error, const char *src, int color);
+char *json_error_get_arrow(json_error_t *error, const char *src, int length, int color);
+char *json_error_get_source_text(json_error_t *error, const char *src);
```
akheron: I'm not sure whether all of these need to be a part of the public API. May we should start by exposing `json_error_get_detailed` and a function that prints directly to the given stream (stdout, stderr, or user-supplied FILE pointer).

Minimizing the public API makes maintaining easier. With more public functions you get more bugs reported :) Other functions can be added to the public API later if needed.

src/error.c
```diff
@@ -61,3 +61,135 @@ void jsonp_error_vset(json_error_t *error, int line, int column,
...
+
+    if (strlen(src) < 2) {
+        return strdup("");
+    }
+
+    if (length < 0) {
+        length = ARROWLEN;
+    }
+
+    if (length >= sizeof(padchars)) {
```
akheron: Travis build failed because here's a comparison between signed and unsigned.

src/error.c
JoakimSoderberg: This is the format Clang uses for it's friendly errors (a lot more friendly than gcc).

Interesting talk about this: https://www.youtube.com/watch?v=NURiiQatBXA

JoakimSoderberg:
Hmm so I guess assuming that `column` was bytes, not utf-8 code points was dumb :)
`int column
The column on which the error occurred. Note that this is the character column, not the byte column, i.e. a multibyte UTF-8 character counts as one column.`

akheron:
Hmm. Even I didn't remember very well, then :)

But if the `column` already is characters and the console supports UTF-8, should you just print an arrow whose length is the same as `column`? For example U+00E4 (ä) is `\xc3\xa4` when encoded as UTF-8, but (per Jansson's documentation) the column only increments by one in this case.

src/error.c
```diff
@@ -61,3 +62,202 @@ void jsonp_error_vset(json_error_t *error, int line, int column,
...
+
+static char *json_error_get_source_text(json_error_t *error, const char *src)
+{
+    const char *start;
+    const char *end;
+    size_t len;
+    char *s;
+
+    // TODO: Pick start properly so we don't split a UTF-8 code point.
+    start = &src[(error->position - error->column)];
```
akheron: This also expects that `column` is bytes, right?

akheron:
What's the status of this? IIRC there were some problems with characters vs. bytes which are not very easy to get right, am I correct?

JoakimSoderberg:
Ah yes, sorry... I kind of lost the steam, my focus has been on this lately: https://github.com/JoakimSoderberg/cargo
Inspired by python argparse :) 

I will try to get to this eventually.

akheron:
Ok, no hurry. cargo looks cool :)

MozLe:
I'd appreciate this too, using JSON for config files means hand editing and more error info makes that much easier.

Zer0-One:
@JoakimSoderberg any chance we might see this PR resolved soon? Like MozLe above, I'm also using JSON for config files, and this would be very useful.

PR #226, opened 2015-03-05
Added support for passing a user argument to custom memory allocation functions

claudix:
I have added support for passing a user argument to custom memory allocation functions. In particular:

- Two additional types have been created: **json_malloc_arg_t** and **json_free_arg_t**, which are similar to their **json_malloc_t** and **json_free_t** counterparts but the former ones get an extra argument (the user argument).

- A new function has been defined: **json_set_alloc_funcs_arg**. It does the same as its counterpart **json_set_alloc_funcs** but also accepts the user argument that will be passed to custom memory allocation functions. 

Passing NULL as the user argument of **json_set_alloc_funcs_arg** treats the specified functions as user argument-less.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -17,15 +17,26 @@
 #undef free
 
 /* memory function pointers */
-static json_malloc_t do_malloc = malloc;
-static json_free_t do_free = free;
+static union { 
+  json_malloc_t     normal;
+  json_malloc_arg_t with_arg;
+} do_malloc = {.normal = malloc};
+
+static union {
+  json_free_t     normal;
+  json_free_arg_t with_arg;
+} do_free = {.normal = free};
+
+/* Argument to pass to alloc functions which accept user argument. If NULL, normal alloc functions are used. */
+static void *user_arg = NULL;
+
 
 void *jsonp_malloc(size_t size)
 {
     if(!size)
         return NULL;
-
-    return (*do_malloc)(size);
+    
+    return (user_arg == NULL) ? (do_malloc.normal)(size) : (do_malloc.with_arg)(size, user_arg);
 }
 
 void jsonp_free(void *ptr)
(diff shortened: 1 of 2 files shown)


harkue:
LGTM.  It would be better to add some example about usage.

src/memory.c
```diff
@@ -56,6 +71,19 @@ char *jsonp_strndup(const char *str, size_t len)
...
-    do_free = free_fn;
+    do_malloc.normal = malloc_fn;
+    do_free.normal = free_fn;
+    user_arg = NULL; 
 }
+
+void json_set_alloc_funcs_arg(json_malloc_arg_t malloc_fn, json_free_arg_t free_fn, void *arg)
+{
+    if (arg == NULL) {
+        json_set_alloc_funcs((json_malloc_t) malloc_fn, (json_free_t) free_fn);
```
ploxiln: Calling `json_set_alloc_funcs_arg()` with a `NULL` arg is most likely a mistake. If the passed malloc/free functions do not actually take an extra arg, then the user calling this variant would cause compiler warnings. If the passed malloc/free functions do take an extra arg and make it this far without compiler warnings, then calling them without the extra arg causes unexpected values from the stack or registers to be interpreted as the extra arg. (I think that for many architectures / calling-conventions, it works fine to call with an extra `NULL` argument that is not expected, it will just be ignored - but it's a mess either way.)

coreyfarrell: My preference is to either assert or have separate storage for the `normal` vs `with_arg` functions.  If someone uses `json_set_alloc_funcs_arg` then we must not call the passed function without the arg.  For separate function storage I think `json_set_alloc_funcs_arg` would need to NULL out the `normal` functions and `json_set_alloc_funcs` would need to NULL out the `with_arg` functions so only the last functions set could take effect.

PR #233, opened 2015-04-30
use snprintf and strncpy in place of sprintf and strcpy

haldean:
This is both good practice and nice for OpenBSD users, who will no
longer get the nag message to not use sprintf/strcpy every time they
link against jansson. It's worth noting that the existing code seems
safe to me - bounds checks were already happening before the actual
calls - and that this is for extra security.


akheron:
Merged, thanks!

haldean:
Thanks for a great library!

PR #234, opened 2015-05-01
Optimized dump_indent to reduce the number of fwrite calls.

flok99:
Previous version would invoke an fwrite for each depth-level. This version tries to minimize that.

### DIFF
--- a/src/dump.c
+++ b/src/dump.c
@@ -50,15 +50,19 @@ static int dump_indent(size_t flags, int depth, int space, json_dump_callback_t 
 {
     if(FLAGS_TO_INDENT(flags) > 0)
     {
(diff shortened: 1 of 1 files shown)


akheron:
Merged, thanks!

flok99:
@akheron I wonder: wouldn't it help to keep a bucket sorted and then use a binary search to find keys in it?

akheron:
If you're talking about the hashtable implementation, then probably no. Each buckets is meant to have at most one item on average, and when the ratio of one item per bucket is exceeded, the hashtable is rehashed with twice the amount of buckets.

Furthermore, the lookup3 hashing algorithm should have a good distribution to avoid hash collisions, and because of hash seeding, malicious inputs that would degrade all items to a single bucket are not easy to construct.

PR #246, opened 2015-07-07
Added "flags" field and getter/setter functions for "flags" field

michael-chen-akamai:
Added field of type "unsigned char" with name "flags" along with json_get_flags() and json_set_flags() getter/setter functions to be used by users of libjansson for saving per node flags that are opaque to libjansson itself.


akheron:
I don't think this is a generally useful feature. The type of the flags is also arguable. Should it be an integer or a void pointer for example.

PR #248, opened 2015-08-13
Update test_memory_funcs.c

phoenix0428:
cannot test "test_memory_funcs.c" because its sample format missed "}"


akheron:
Merged, thanks.

PR #252, opened 2015-08-31
Buildable with MS Compiler version  >= 1900

OlehKulykov:


OlehKulykov:
With the current project structure *snprintf* can be defined twice: in files *jansson_private_config.h.cmake*, *jansson_private.h*.
Anyway, with *MS Compiler version >= 1900* snprintf should not be a macro definition, so tracking & availability of this one was carefully added to both this files.

akheron:
Merged, thanks.

PR #253, opened 2015-09-10
FIX CMake snprintf DETECTION

BurningEnlightenment:
Replace check_function_exists calls with check_symbol_exists calls for
snprintf detection because starting with MSVC 19 snprintf is declared as
an inline function thus check_function_exists fails to detect it.
(See http://public.kitware.com/Bug/view.php?id=15659)


akheron:
I believe this was fixed by 8f2298bad8f77a22efe0dc9a95676fae6c203e36

PR #255, opened 2015-10-05
Improve usage of snprintf in dump.c

gonzzor:
Be more explicit about casting arguments to snprintf.

Use sizeof() instead of magic constant.


akheron:
Merged, thanks.

PR #257, opened 2015-10-12
changed code from while to do/while

denis2342:


akheron:
There are a couple of other such loops in the same file. If you change one of them, the others should also be changed.

denis2342:
ok, I changed the other four too.

akheron:
Merged, thanks!

PR #263, opened 2015-12-18
pack_unpack: List first unrecognized key in strict unpacking

wking:
Otherwise figuring out what's wrong with your JSON can be tricky,
especially if you're using a single fmt string to validate a large,
complicated schema.

This just shows the first unrecognized key, because assembling a list
of all unrecognized strings seemed like too much work.  Humans who are
supplying invalid JSON can always remove the listed key and
re-validate.  Programs seeking an API listing unrecognized keys are
out of luck.

### DIFF
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -436,6 +436,8 @@ static int unpack_object(scanner_t *s, json_t *root, va_list *ap)
     if(root && strict == 1) {
         /* We need to check that all non optional items have been parsed */
         const char *key;
+        int have_unrecognized_keys = 0;
+        strbuffer_t unrecognized_keys;
         json_t *value;
         long unpacked = 0;
         if (gotopt) {
@@ -443,6 +445,14 @@ static int unpack_object(scanner_t *s, json_t *root, va_list *ap)
             json_object_foreach(root, key, value) {
                 if(!hashtable_get(&key_set, key)) {
                     unpacked++;
+
+                    if (!have_unrecognized_keys) {
+                        strbuffer_init(&unrecognized_keys);
+                        have_unrecognized_keys = 1;
+                    } else {
+                        strbuffer_append_bytes(&unrecognized_keys, ", ", 2);
+                    }
+                    strbuffer_append_bytes(&unrecognized_keys, key, strlen(key));
                 }
             }
         } else {
@@ -450,7 +460,24 @@ static int unpack_object(scanner_t *s, json_t *root, va_list *ap)
             unpacked = (long)json_object_size(root) - (long)key_set.size;
         }
         if (unpacked) {
-            set_error(s, "<validation>", "%li object item(s) left unpacked", unpacked);
+            if (!gotopt) {
+                /* Find the first unrecognized key */
+                json_object_foreach(root, key, value) {
+                    if(!hashtable_get(&key_set, key)) {
+                        if (!have_unrecognized_keys) {
+                            strbuffer_init(&unrecognized_keys);
+                            have_unrecognized_keys = 1;
+                        } else {
+                            strbuffer_append_bytes(&unrecognized_keys, ", ", 2);
+                        }
+                        strbuffer_append_bytes(&unrecognized_keys, key, strlen(key));
+                    }
+                }
+            }
+            set_error(s, "<validation>",
+                      "%li object item(s) left unpacked: %s",
+                      unpacked, strbuffer_value(&unrecognized_keys));
+            strbuffer_close(&unrecognized_keys);
             goto out;
         }
     }
--- a/test/suites/api/test_unpack.c
+++ b/test/suites/api/test_unpack.c
@@ -298,10 +298,16 @@ static void run_tests()
     json_decref(j);
 
     /* Unpack the same item twice */
-    j = json_pack("{s:s, s:i}", "foo", "bar", "baz", 42);
+    j = json_pack("{s:s, s:i, s:b}", "foo", "bar", "baz", 42, "quux", 1);
     if(!json_unpack_ex(j, &error, 0, "{s:s,s:s!}", "foo", &s, "foo", &s))
         fail("json_unpack object with strict validation failed");
-    check_error("1 object item(s) left unpacked", "<validation>", 1, 10, 10);
+    {
+        const char *possible_errors[] = {
+            "2 object item(s) left unpacked: baz, quux",
+            "2 object item(s) left unpacked: quux, baz"
+        };
+        check_errors(possible_errors, 2, "<validation>", 1, 10, 10);
+    }
     json_decref(j);
 
     j = json_pack("[i,{s:i,s:n},[i,i]]", 1, "foo", 2, "bar", 3, 4);
@@ -335,7 +341,7 @@ static void run_tests()
     j = json_pack("{s{snsn}}", "foo", "bar", "baz");
     if(!json_unpack_ex(j, &error, 0, "{s{sn!}}", "foo", "bar"))
         fail("json_unpack nested object with strict validation failed");
-    check_error("1 object item(s) left unpacked", "<validation>", 1, 7, 7);
+    check_error("1 object item(s) left unpacked: baz", "<validation>", 1, 7, 7);
     json_decref(j);
 
     /* Error in nested array */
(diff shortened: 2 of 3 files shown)


akheron:
This is an interesting patch and I believe it would be useful to many users.

Using `strbuffer_t` in `src/strbuffer.h` it would be quite easy to list all unrecognized keys. Testing it is a bit harder, though, as you cannot depend on getting the keys in the same order every time.

akheron:
Here's a quick patch on top of your changes that implements listing all unrecognized keys. I'd appreciate if you incorporate it into your branch, check that it's a sane thing to do, and perhaps add some more tests.

```diff
diff --git a/src/pack_unpack.c b/src/pack_unpack.c
index d2118cf..617b3c0 100644
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -435,7 +435,9 @@ static int unpack_object(scanner_t *s, json_t *root, va_list *ap)
 
     if(root && strict == 1) {
         /* We need to check that all non optional items have been parsed */
-        const char *key, *unrecognized_key;
+        const char *key;
+        int have_unrecognized_keys = 0;
+        strbuffer_t unrecognized_keys;
         json_t *value;
         long unpacked = 0;
         if (gotopt) {
@@ -443,7 +445,14 @@ static int unpack_object(scanner_t *s, json_t *root, va_list *ap)
             json_object_foreach(root, key, value) {
                 if(!hashtable_get(&key_set, key)) {
                     unpacked++;
-                    unrecognized_key = key;
+
+                    if (!have_unrecognized_keys) {
+                        strbuffer_init(&unrecognized_keys);
+                        have_unrecognized_keys = 1;
+                    } else {
+                        strbuffer_append_bytes(&unrecognized_keys, ", ", 2);
+                    }
+                    strbuffer_append_bytes(&unrecognized_keys, key, strlen(key));
                 }
             }
         } else {
@@ -455,14 +464,20 @@ static int unpack_object(scanner_t *s, json_t *root, va_list *ap)
                 /* Find the first unrecognized key */
                 json_object_foreach(root, key, value) {
                     if(!hashtable_get(&key_set, key)) {
-                        unrecognized_key = key;
-                        break;
+                        if (!have_unrecognized_keys) {
+                            strbuffer_init(&unrecognized_keys);
+                            have_unrecognized_keys = 1;
+                        } else {
+                            strbuffer_append_bytes(&unrecognized_keys, ", ", 2);
+                        }
+                        strbuffer_append_bytes(&unrecognized_keys, key, strlen(key));
                     }
                 }
             }
             set_error(s, "<validation>",
-                      "%li object item(s) left unpacked, including %s",
-                      unpacked, unrecognized_key);
+                      "%li object item(s) left unpacked: %s",
+                      unpacked, strbuffer_value(&unrecognized_keys));
+            strbuffer_close(&unrecognized_keys);
             goto out;
         }
     }
diff --git a/test/suites/api/test_unpack.c b/test/suites/api/test_unpack.c
index 87caaf6..babe0dd 100644
--- a/test/suites/api/test_unpack.c
+++ b/test/suites/api/test_unpack.c
@@ -298,10 +298,16 @@ static void run_tests()
     json_decref(j);
 
     /* Unpack the same item twice */
-    j = json_pack("{s:s, s:i}", "foo", "bar", "baz", 42);
+    j = json_pack("{s:s, s:i, s:b}", "foo", "bar", "baz", 42, "quux", 1);
     if(!json_unpack_ex(j, &error, 0, "{s:s,s:s!}", "foo", &s, "foo", &s))
         fail("json_unpack object with strict validation failed");
-    check_error("1 object item(s) left unpacked, including baz", "<validation>", 1, 10, 10);
+    {
+        const char *possible_errors[] = {
+            "2 object item(s) left unpacked: baz, quux",
+            "2 object item(s) left unpacked: quux, baz"
+        };
+        check_errors(possible_errors, 2, "<validation>", 1, 10, 10);
+    }
     json_decref(j);
 
     j = json_pack("[i,{s:i,s:n},[i,i]]", 1, "foo", 2, "bar", 3, 4);
@@ -335,7 +341,7 @@ static void run_tests()
     j = json_pack("{s{snsn}}", "foo", "bar", "baz");
     if(!json_unpack_ex(j, &error, 0, "{s{sn!}}", "foo", "bar"))
         fail("json_unpack nested object with strict validation failed");
-    check_error("1 object item(s) left unpacked, including baz", "<validation>", 1, 7, 7);
+    check_error("1 object item(s) left unpacked: baz", "<validation>", 1, 7, 7);
     json_decref(j);
 
     /* Error in nested array */
@@ -395,6 +401,6 @@ static void run_tests()
     i1 = i2 = i3 = 0;
     if(!json_unpack_ex(j, &error, 0, "{sis?i!}", "foo", &i1, "bar", &i2))
         fail("json_unpack failed for optional values with strict mode and compensation");
-    check_error("1 object item(s) left unpacked, including baz", "<validation>", 1, 8, 8);
+    check_error("1 object item(s) left unpacked: baz", "<validation>", 1, 8, 8);
     json_decref(j);
 }
diff --git a/test/suites/api/util.h b/test/suites/api/util.h
index 1bc4c9c..c25f4eb 100644
--- a/test/suites/api/util.h
+++ b/test/suites/api/util.h
@@ -61,6 +61,45 @@
     } while(0)
 
 
+/* Assumes json_error_t error */
+#define check_errors(texts_, num_, source_, line_, column_, position_)  \
+    do {                                                                \
+        int i_, found_ = 0;                                             \
+        for(i_ = 0; i_ < num_; i_++) {                                  \
+            if(strcmp(error.text, texts_[i_]) == 0) {                   \
+                found_ = 1;                                             \
+                break;                                                  \
+            }                                                           \
+        }                                                               \
+        if (!found_) {                                                  \
+            failhdr;                                                    \
+            fprintf(stderr, "text: \"%s\" does not match\n", error.text); \
+            exit(1);                                                    \
+        }                                                               \
+        if(strcmp(error.source, source_) != 0) {                        \
+            failhdr;                                                    \
+                                                                        \
+            fprintf(stderr, "source: \"%s\" != \"%s\"\n", error.source, source_); \
+            exit(1);                                                    \
+        }                                                               \
+        if(error.line != line_) {                                       \
+            failhdr;                                                    \
+            fprintf(stderr, "line: %d != %d\n", error.line, line_);     \
+            exit(1);                                                    \
+        }                                                               \
+        if(error.column != column_) {                                   \
+            failhdr;                                                    \
+            fprintf(stderr, "column: %d != %d\n", error.column, column_); \
+            exit(1);                                                    \
+        }                                                               \
+        if(error.position != position_) {                               \
+            failhdr;                                                    \
+            fprintf(stderr, "position: %d != %d\n", error.position, position_); \
+            exit(1);                                                    \
+        }                                                               \
+    } while(0)
+
+
 static void run_tests();
 
 int main() {

wking:
On Mon, Dec 21, 2015 at 12:14:28AM -0800, Petri Lehtinen wrote:
> Here's a quick patch on top of your changes that implements listing
> all unrecognized keys. I'd appreciate if you incorporate it into
> your branch, check that it's a sane thing to do, and perhaps add
> some more tests.

Incorporated with e444516 → bc0f787.  The only change I made vs. your
patch was to redefine check_error in terms of the more general
check_errors to keep things DRYer.  In the new commit message, I point
out possible confusion with comma-containing keys (which I don't think
we need to worry about, since I haven't seen such keys in the wild)
and my preference for using unrecognized_keys.value == NULL instead of
a separate have_unrecognized_keys.  I'm happy to reroll to use value
== NULL if that approach sounds acceptable to you.

wking:
On Mon, Dec 21, 2015 at 12:14:28AM -0800, Petri Lehtinen wrote:
> … and perhaps add some more tests.

And the current coverage seems sufficient to me.  If you have any
use-cases you feel need more testing, just let me know what they are.

akheron:
`strbuffer_close` isn't called, and that causes the memory leak seen by the CI tests.

wking:
On Tue, Dec 22, 2015 at 03:00:40AM -0800, Petri Lehtinen wrote:
> `strbuffer_close` isn't called, and that causes the memory leak seen
> by the CI tests.

Ah, that was another change I made vs. your suggested patch :p.  In
hindsight, needing a separate free after strbuffer_steal_value makes
sense.  Fixed in bc0f787 → 1d513d0.

akheron:
Merged, thanks a lot!

PR #264, opened 2015-12-21
Add json_get_alloc_funcs() to allow alloc function fetching

npmccallum:
This is particularly useful in modular situations where the allocation
functions are either unknown or private. For instance, in such cases,
the caller of json_dumps() has no way to free the returned buffer.


akheron:
Merged, thanks!

PR #265, opened 2015-12-30
Add json_search function, that search if a json element exists within another json element

babelouest:
I have this need for my own purpose, but I thought it could be useful for others.


akheron:
Hi and sorry for the late reply. I don't think this function is of enough general usefulness that it could be added to Jansson.

babelouest:
No problem, thanks

PR #267, opened 2016-02-03
Optimized functions

Mephistophiles:
Optimized json_equal, json_delete, json_copy, json_deep_copy
Replaced if to switch
Removed unnecessary check for NULL.


akheron:
Merged, thanks.

PR #268, opened 2016-02-10
Fix libdir is wrong path when compile with cmake

wingyplus:
`pkg-config --libs jansson` is wrong when compiling with cmake

```bash
$ pkg-config --libs jansson
-L/usr/local//usr/local/lib -ljansson
```

because `JANSSON_INSTALL_LIB_DIR` is modified after foreach block in https://github.com/akheron/jansson/blob/master/CMakeLists.txt#L622

to fix this, I move fill configuration jansson.pc.in before L622 and remove some duplicate.


akheron:
Merged, thanks!

PR #269, opened 2016-02-11
Small fixes

Mephistophiles:
* Fixed uninit variables
* Removed check less 0 with unsigned


src/pack_unpack.c
```diff
@@ -440,6 +440,7 @@ static int unpack_object(scanner_t *s, json_t *root, va_list *ap)
         strbuffer_t unrecognized_keys;
         json_t *value;
         long unpacked = 0;
+        strbuffer_init(&unrecognized_keys);
```
akheron: It's a performance optimization to initialize `unrecognized_keys` only if it's needed.

Mephistophiles: Okey, I removed from his commit this place.

PR #273, opened 2016-02-27
Add VAPI in order to support Vala language

mspanc:
I need to use Jansson from the [Vala](http://live.gnome.org/Vala) programming language. This requires to create thin binding layer called VAPI that tells Vala compiler how to translate its code into corresponding C code.

I made initial version of the VAPI. It is 90% done, I only omitted a few functions that need more work. I will commit more updates soon but so far this PR should mostly indicate that this is Work In Progress.

Once it's ready to merge, I need your review as I don't use CMake and I don't know yet how to make VAPI ship with the project. It should land into $PREFIX/share/vala/jansson.vapi (and should be packaged to -dev packages by OS maintainers) in order to be recognized by the Vala compiler.

### DIFF
--- a/jansson.vapi
+++ b/jansson.vapi
@@ -1,0 +1,239 @@
+[CCode (cheader_filename = "jansson.h", lower_case_cprefix = "json_")]
+namespace Jansson {
+  [CCode (cname = "JANSSON_MAJOR_VERSION")]
+  public static int MAJOR_VERSION;
+
+  [CCode (cname = "JANSSON_MINOR_VERSION")]
+  public static int MINOR_VERSION;
+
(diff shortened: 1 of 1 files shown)


akheron:
Is it common for projects like Jansson to ship a vala binding or to have it as a separate project? Do you have examples of other libraries shipping it?

mspanc:
Yes it is common. See libpulse

akheron:
To me, Autoconf/Automake is the "official" way to build source tarballs and install files. If you add jansson.vapi  to be installed to `$datadir/vala` via Automake, we can find out the right way to do this in CMake later.

PR #278, opened 2016-03-22
add json_object_update_missing_incref

dancal:
add function

json_object_update_missing_incref


akheron:
json_object_set_nocheck increments the reference count so this is unnecessary.

PR #279, opened 2016-03-25
add auto_json_decref macro

sjaeckel:
gcc and clang implement the compiler attribute cleanup() [1] which lets
you add a function call that is invoked as soon as a variable goes out
of scope.

This feature improves code where you retrieve a json_t object at one
point and you're not interested to keep it over a longer period of time.

[1] https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-g_t_0040code_007bcleanup_007d-variable-attribute-3485

### DIFF
--- a/src/jansson.h
+++ b/src/jansson.h
@@ -112,6 +112,20 @@ void json_decref(json_t *json)
         json_delete(json);
 }
 
+#ifdef __GNUC__
+#define auto_json_decref __attribute((cleanup(json_autodecref)))
+static JSON_INLINE
+void json_autodecref(json_t **json)
+{
+    if(!json)
+        return;
+    if(!*json)
+        return;
+    json_decref(*json);
+    *json = NULL;
+}
+#endif
+
 
 /* error reporting */
 
--- a/test/suites/api/test_simple.c
+++ b/test/suites/api/test_simple.c
@@ -9,6 +9,37 @@
 #include <jansson.h>
 #include "util.h"
 
+#if defined __GNUC__
+static int check_free_calls;
+static void check_free(void *p)
+{
+    check_free_calls++;
+    if(p) free(p);
+}
+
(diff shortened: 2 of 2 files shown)


Mephistophiles:
Home immediately know all the dangerous moments, such as:

```c
void blah (json_t * test) {
	auto_json_decref json_t * str = json_string ( "test");

	json_object_set_new_nocheck (test, "blah", str);
}
```

sjaeckel:
@Mephistophiles as already said in the commit "...where you retrieve a json_t object at one
point and you're not interested to keep it over a longer period of time." and if you know that `json_object_set_new_nocheck()` takes ownership of `str` why would you add a `auto_json_decref`?

I just want to say: people who'd add a `auto_json_decref` there, would also just do `json_decref(str)`

sjaeckel:
@akheron is there a chance this will be merged?

akheron:
Yes, there definitely is a chance. My only concern is that this cannot be used in portable code, and that has to be communicated to users very clearly.

It's a shame actually. If all compilers had a way to do this, this should be the default and you would need to use an attribute to disable autodecref.

sjaeckel:
> My only concern is that this cannot be used in portable code...

That's for sure :-( Therefor it's also enclosed in `#ifdef __GNUC__` ...

> ... and that has to be communicated to users very clearly.

Can I help you with anything there?

> It's a shame actually. If all compilers had a way to do this, this should be the default and you would need to use an attribute to disable autodecref.

Yes it's a shame, but who uses a compiler that isn't based on or at least supports most of the functionality of gcc anyways? ...icc supports it and who cares about msvc... ;-)

akheron:
> Can I help you with anything there?

You could document it. That's what has to be done in any case :)

akheron:
Superseded by #301

PR #295, opened 2016-06-27
WIP/RFC: set errno to avoid return value ambiguity

Thynix:
In some situations, such as `json_object_size()`, `json_object_get()`, and
`json_string_length()`, it is ambiguous whether a return value indicates an
error. Setting errno on error is one way to avoid this.

Is this a direction worth pursuing? This already demonstrates the general shape
of approach I'd take: return specific error codes from internal functions, set
errno to them, and leave the external functions' return values unchanged. This
would open up the option of documenting possible errors in more detail.

### DIFF
--- a/doc/apiref.rst
+++ b/doc/apiref.rst
@@ -581,27 +581,27 @@ allowed in object keys.
 
    .. refcounting:: new
 
-   Returns a new JSON object, or *NULL* on error. Initially, the
-   object is empty.
+   Returns a new JSON object, or *NULL* on memory allocation failure. Initially,
+   the object is empty.
 
 .. function:: size_t json_object_size(const json_t *object)
 
-   Returns the number of elements in *object*, or 0 if *object* is not
-   a JSON object.
+   Returns the number of elements in *object*, or returns 0 and sets *errno* if
+   *object* is not a JSON object.
 
 .. function:: json_t *json_object_get(const json_t *object, const char *key)
 
    .. refcounting:: borrow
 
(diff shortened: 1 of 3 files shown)


akheron:
I perceive errno as a way for standard library functions to communicate various error codes, and not generally available to other libraries. The C89 and C99 standards only define `EDOM`, `EILSEQ` and `ERANGE`, which are not very useful by themselves. So I don't think this is the correct approach.

A note on the patch: A library function must never set errno to zero. Your patch does this in many places.

Thynix:
Do you have suggestions for resolving these return value ambiguities? If
errno is not an option (Is there documentation you can point to about errno
being only for standard library functions' use? I did find
http://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html#tag_15_03
about not zeroing it.) two things I can think of are:

* Break API and change the return values, possibly adding pointer arguments
in cases where all return values are valid. I'm not keen on breaking API or
the messiness of error arguments for basic failure information.
* Add a thread-local Jansson-specific errno equivalent. This seems... odd,
but should avoid the conflict you mention.

akheron:
The third option is to do nothing. The functions you listed only return with an error if a value of an invalid type is passed. If you're not sure about what you're passing as an argument, check the type before the call.

Functions that create new values return NULL on error, and the only error situation is out of memory.

Decoding has its own `json_error_t` to communicate errors back to the user.

AFAICS, the only operation that has multiple possible failure scenarios is encoding to a file where either out of memory or an I/O error can happen. But if it's an I/O error, it's very likely that errno is available.

PR #298, opened 2016-08-11
OOM crash 

pasiopou:
I have a case with a crash after OOM due to `hashtable->buckets` being `NULL` in `hashtable_set()`. 
Here is a test case which reproduces it and a proposed fix.


akheron:
Merged, thanks!

PR #305, opened 2016-09-11
Allow install targets to be optionally defined

warmwaffles:
This is nice to have if you build this project from source within a project and link statically.


warmwaffles:
heh, makes no sense why coverage decreased 😉

akheron:
Merged, thanks!

PR #309, opened 2016-10-02
Allow parsing of buffers larger than 2GB on most 64 bit arch.

halfaleague:
size_t is usually 64 bits on most architectures -- this allows for larger .json files


akheron:
Merged, thanks!

PR #310, opened 2016-10-05
Add configure script

owtaylor:
Continuous integration and other mass-build tasks can be simplified if projects have a uniform way to be compiled without a lot of per-project setup. See http://people.gnome.org/~walters/docs/build-api.txt for details. This PR adds a wrapper configure script that accepts common options like --prefix and --libdir and runs cmake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ config.h.in
 config.log
 config.status
 config.sub
-configure
 depcomp
 install-sh
 libtool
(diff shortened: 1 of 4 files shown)


akheron:
The release tarballs already include a configure script (via autotools).

owtaylor:
I'm embarrassed to admit I didn't notice the dual CMake and autotools build setup. Assuming that the plan is to support both long-term, I'll do another PR to add an 'autogen.sh' script (see the builld-api.txt document referenced above) that runs autoreconf.

akheron:
autogen.sh sounds perfect :)

PR #311, opened 2016-10-07
Fix passing redundant argument to error_set()

robertmu:
Passing redundant argument to error_set(), should be remove:)


akheron:
Merged, thanks!

PR #312, opened 2016-10-16
Homogenizing includes.

jspark311:
These changes were required to build with a private config. Resulting build tests ok for me.

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -5,9 +5,7 @@
  * it under the terms of the MIT license. See LICENSE for details.
  */
 
-#if HAVE_CONFIG_H
-#include <jansson_private_config.h>
-#endif
+#include "jansson.h"   /* for JSON_INLINE */
 
 #include <stdlib.h>
 #include <string.h>
@@ -16,7 +14,6 @@
 #include <stdint.h>
 #endif
 
-#include <jansson_config.h>   /* for JSON_INLINE */
 #include "jansson_private.h"  /* for container_of() */
 #include "hashtable.h"
 
--- a/src/hashtable_seed.c
+++ b/src/hashtable_seed.c
@@ -3,7 +3,7 @@
 */
 
 #ifdef HAVE_CONFIG_H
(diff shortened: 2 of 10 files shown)


jspark311:
Failed automated tests. Addressing.....

jspark311:
Ok... I found my oversight (maybe?).
Reading the build documentation, I gathered that the include option -DHAVE_CONFIG_H should cause a mutually-exclusive choice between jansson_config.h and jansson_private_config.h. Is this the case?

If so, the build-apparatus surrounding config generation is not complete, as JSON_INLINE is not defined in the generated jansson_private_config.h file.

My (working) private config file contains this:

    #ifndef __PRIVATE_JANSSON_CONFIG_H__
    #define __PRIVATE_JANSSON_CONFIG_H__
    #define JSON_PARSER_MAX_DEPTH   2048
    #define JSON_INLINE           inline
    #define JSON_HAVE_LOCALECONV       0
    #define JSON_INTEGER_IS_LONG_LONG  0
    #define HAVE_UNISTD_H              1
    #define HAVE_SYS_TYPES_H           1
    #define HAVE_INTTYPES_H            1
    #define HAVE_STDINT_H              1
    #endif   //__PRIVATE_JANSSON_CONFIG_H__

Any insight?

akheron:
The meaning of the two header files is this:
- `jansson_private_config.h` is used when compiling Jansson itself (FWIW, it was formerly called `config.h` which is a standard name in the autoconf world, but renamed for easier CMake integration). 
- `jansson_config.h` is used when someone uses Jansson in his or her own program, i.e. it's installed to `/usr/include` or equivalent.

This has the implication that `jansson_private_config.h` cannot be referenced in `jansson.h` or `jansson_config.h` because it's not available anymore after Jansson itself has been compiled to a library.

Both files are either generated by `./configure` or `cmake`, or by hand if you use neither of them.

jspark311:
That makes things make more sense. I must hand-generate jansson_config.h as well.
Please disregard PR.

PR #314, opened 2016-10-19
part of conditional expression is always true (!done)

martinlindhe:
pvs-studio reports this issue as

```
src/hashtable_seed.c<#~>error<#~>V560<#~>A part of conditional expression is always true: 
	!done.<#~>false<#~>2<#~>#if !defined(_WIN32) && defined(USE_URANDOM)<#~>
    if (!done && seed_from_urandom(&seed) == 0)<#~>        done = 1;<#~>
```


akheron:
Merged, thanks!

PR #321, opened 2016-11-13
parens for LONG_LONG_INT in cmake

brian-armstrong:
when this is defined as "", cmake bails with an error about arguments in
if


akheron:
Merged, thanks!

PR #323, opened 2016-11-21
Add json_array_contains

Mephistophiles:
These functions are useful for storing unique data.

Co-Authored-By: overcrook <simple-one@mail.ru>


doc/apiref.rst
```diff
@@ -521,6 +525,18 @@ A JSON array is an ordered collection of other JSON values.
    *value*. This is useful when *value* is newly created and not used
    after the call.
 
+.. function:: int json_array_append_unique(json_t *array, json_t *value)
+
+   Appends *value* to the end of *array*, but only if *array* does not contain
+   *value* unique. Returns 0 on success and -1 on error.
```
Zer0-One: Grammar mistake, should be something like "but only if *array* does not already contain *value*."

akheron:
Sorry for the late response.

The `json_array_contains()` function seems useful to me. The documentation could state, however, that it uses `json_equal()` to compare two values. It could also take a pointer to a comparator function, and use `json_equal` if the provided comparator is NULL.

The `json_array_append_unique` and `json_array_append_unique_new` don't seem like very widely useful functions to me.

Mephistophiles:
Updated. json_array_append_unique and json_array_append_unique_new were removed from patch.

doc/apiref.rst
```diff
@@ -497,6 +497,11 @@ A JSON array is an ordered collection of other JSON values.
    if *array* is *NULL*, or if *index* is out of range, *NULL* is
    returned.
 
+.. function:: json_t *json_array_contains(const json_t *array, json_t *value)
```
akheron: The signature is wrong

doc/apiref.rst
```diff
@@ -497,6 +497,11 @@ A JSON array is an ordered collection of other JSON values.
...
+
+   Returns 1 if *array* contains *value* and 0 otherwise. That it uses
+   `json_equal()` to compare two values.
+
```
akheron: Please add `.. versionadded:: 2.10` here, because 2.10 will be the first version that has this function.

doc/apiref.rst
Mephistophiles: Fixed

doc/apiref.rst
Mephistophiles: Fixed

Mephistophiles:
Updated

doc/apiref.rst
akheron: It's still wrong...

PR #329, opened 2017-02-26
Add JSON_EMBED encoding flag

npmccallum:
The JSON_EMBED encoding flag causes the opening and closing characters
of the top-level array ('[', ']') or object ('{', '}') to be omitted
during encoding. This feature makes it possible to concatenate multiple
arrays or objects in the stream output. It also makes it possible to
perform outputs of partial composes.

One such example of a partial compose is when outputting a JWE object.
The output is a JSON object. But it has one top-level attribute
("ciphertext") that can grow out of proportion with the rest of the
metadata. With the JSON_EMBED flag, the other metadata can be composed
ahead of time and dumped during the beginning of output, where the
"ciphertext" and "tag" attributes can be streamed out in chunks. Thus,
the header material can be composed with Jansson and the ciphertext
itself can be composed manually.

### DIFF
--- a/doc/apiref.rst
+++ b/doc/apiref.rst
@@ -940,6 +940,13 @@ can be ORed together to obtain *flags*.
 
    .. versionadded:: 2.7
 
+``JSON_EMBED``
+   If this flag is used, the opening and closing characters of the top-level
+   array ('[', ']') or object ('{', '}') are omitted during encoding. This
+   flag is useful when concatenating multiple arrays or objects into a stream.
+
+   .. versionadded:: 2.10
+
 These functions output UTF-8:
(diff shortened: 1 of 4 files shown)


npmccallum:
These tests appear to have failed due to a network connectivity issue with github.com. Can you please re-run them manually?

npmccallum:
I was able to fix tests on windows, but the Travis CI tests are broken due to networking restrictions.

npmccallum:
I'm not sure why, but it now appears that all the Travis CI tests pass too. So this patch should be close.

akheron:
Merged, thanks!

PR #339, opened 2017-04-18
Enable optional object members in json_pack()

npmccallum:
For example, the following call creates an empty object:

  json_pack("{s?s}", "foo", NULL);

### DIFF
--- a/doc/apiref.rst
+++ b/doc/apiref.rst
@@ -1269,6 +1269,14 @@ arguments.
 
     .. versionadded:: 2.8
 
+``s*`` (string) [const char \*]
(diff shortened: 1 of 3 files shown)


akheron:
`json_pack("{s:**}", "a", NULL)` now produces an empty object instead of an error.

I think `json_pack()` needs to somehow make a distinction between the two error cases: that the argument for `s`, `o` or `O` was `NULL`, or that the format character was invalid.

npmccallum:
@akheron I have fixed this problem and added a test so that it doesn't reappear.

akheron:
Merged, thanks a lot!

PR #347, opened 2017-05-23
Tests fail with Russian locale

artdol:


akheron:
Jansson should work with any locale. What error did you get?

coreyfarrell:
@artdol could you let us know what error you get without this change?

PR #350, opened 2017-06-06
Add jansson command

npmccallum:
The jansson command is a utility for building and parsing JSON objects
from the CLI. It is a simple stack machine on which operations can be
performed. This command exposes the most important Jansson APIs.


npmccallum:
I think this is ready for a review. We're still missing a man page. I also wasn't able to figure out how to include cmd/jansson.c in the coveralls check. But before I put the effort into all of this, I want to make sure you're interested in merging this first.

akheron:
I'm not sure this belongs to the Jansson repo, as Jansson is a library.

npmccallum:
The idea is that this is basic JSON parsing. It doesn't require anything except Jansson and users of the Jansson library will probably be highly interested in this functionality. It could become its own project, but this would require people adding a dependency. Alternatively, I could add it in [José](https://github.com/latchset/jose), but then people who want just JSON parsing would have to get a whole crypto stack. So since people who want JSON parsing are already interested in Jansson for C, I thought this would be a relevant place.

Let me know if you don't want it ASAP so I can start merging it somewhere else.

sopvop:
I thought `jq` was standard and feature rich command line tool for JSON.

npmccallum:
I've decided to merge this utility into the José project so that I can integrate directly with some José APIs. Sorry for the bother.

svnpenn:
@npmccallum where is the Jose project

PR #368, opened 2017-10-29
Fix generated `pkg-config` file using cmake.

AmeyaVS:
Fixed the generated `jansson.pc` with cmake to be consistent with the
one generated using GNU Autotools.


akheron:
Merged, thanks!

Issue #135, opened 2013-08-28
Check that all functions that steal a value decref it on error


akheron:

fabled:
This is probably most important for json_pack(). When you have multiple 'o' specifiers, and an error occurs, it's impossible for the caller to know which references where stolen and which not. The function should probably always steal the references (e.g. explicitly call decref on the remainder of 'o' specifiers not already stolen).

akheron:
Good point! Would you like to write a patch?

coreyfarrell:
@fabled I posted a patch for json_pack.  Sorry if this is a duplicate message, I'm not sure if github notifies issue participants of related pull requests.

PR #374, opened 2017-11-07
json_pack: Enable more complete stealing of references.

coreyfarrell:
Users of the "o" format have an expectation that the object reference
will be stolen.  Any error causes the collection process to end early.
This patch causes json_pack and related functions to continue scanning
the format and parameters so all references can be stolen to prevent
leaks.  This makes no attempt to continue processing if the format
string is broken or missing.

'make check' still passes.  Ran test_pack under valgrind and verified
that the leaked reference is fixed.

Issue #135

### DIFF
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -29,6 +29,7 @@ typedef struct {
     int line;
     int column;
     size_t pos;
+    int has_error;
 } scanner_t;
 
 #define token(scanner) ((scanner)->token.token)
@@ -60,6 +61,7 @@ static void scanner_init(scanner_t *s, json_error_t *error,
     s->line = 1;
     s->column = 0;
     s->pos = 0;
+    s->has_error = 0;
 }
 
 static void next_token(scanner_t *s)
@@ -136,6 +138,7 @@ static char *read_string(scanner_t *s, va_list *ap,
     t = token(s);
     prev_token(s);
 
+    *ours = 0;
     if(t != '#' && t != '%' && t != '+') {
         /* Optimize the simple case */
         str = va_arg(*ap, const char *);
@@ -153,7 +156,6 @@ static char *read_string(scanner_t *s, va_list *ap,
         }
 
         *out_len = length;
-        *ours = 0;
         return (char *)str;
     }
 
@@ -163,8 +165,7 @@ static char *read_string(scanner_t *s, va_list *ap,
         str = va_arg(*ap, const char *);
         if(!str) {
             set_error(s, "<args>", json_error_null_value, "NULL string argument");
-            strbuffer_close(&strbuff);
-            return NULL;
+            s->has_error = 1;
         }
 
         next_token(s);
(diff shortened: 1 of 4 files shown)


akheron:
Thanks for the PR and sorry I haven't replied earlier. It seems that GitHub only sends me ~20% of emails I should have, and therefore I easily miss new issues and PRs :/

In general this looks good. I'll add some comments inline.

src/pack_unpack.c
```diff
@@ -29,6 +29,7 @@ typedef struct {
     int line;
     int column;
     size_t pos;
+    int scan_mode;
```
akheron: I don't like the `scan_mode` name. It should reflect the fact that an error was encountered and "error recovery" is in progress.

src/pack_unpack.c
```diff
@@ -368,7 +383,7 @@ static json_t *pack(scanner_t *s, va_list *ap)
                 prev_token(s);
 
             json = va_arg(*ap, json_t *);
-            if (!json && nullable) {
+            if (s->scan_mode || (!json && nullable)) {
```
akheron: Could you explain this? Why a JSON null is returned when recovering from an error?

src/pack_unpack.c
```diff
@@ -798,6 +814,10 @@ json_t *json_vpack_ex(json_error_t *error, size_t flags,
         set_error(&s, "<format>", json_error_invalid_format, "Garbage after format string");
         return NULL;
     }
+    if(s.scan_mode && value) {
```
akheron: The `&& value` is technically not needed, because it's already checked a few lines up.

src/pack_unpack.c
coreyfarrell: `int error_recovery;` ?

src/pack_unpack.c
coreyfarrell: When we are already in error recovery mode it is pointless to `json_incref(json)` just so it has to be unreferenced when we finally destroy the object.  It's likely an insignificant optimization to the error path, I can remove it if you want.

src/pack_unpack.c
coreyfarrell: I've removed the unneeded check here and a few other places where it was not needed.

coreyfarrell:
> Thanks for the PR and sorry I haven't replied earlier. It seems that GitHub only sends me ~20% of emails I should have, and therefore I easily miss new issues and PRs :/

If I have a PR in the future and don't hear from you after a week should I post elsewhere to notify you?

src/pack_unpack.c
akheron: Yeah, or maybe just `has_error`

src/pack_unpack.c
akheron: I think it would be best to remove it, to not be confusing in the future

src/pack_unpack.c
akheron: :+1:

akheron:
> If I have a PR in the future and don't hear from you after a week should I post elsewhere to notify you?

You can email me or send a message to the [jansson-users group](https://groups.google.com/d/forum/jansson-users).

akheron:
Merged, thanks!

PR #381, opened 2017-12-20
Use a more specific error code for premature end of input

phst:


akheron:
Hi and thanks for the pull request. Sorry I didn't notice it earlier.

If you're still interested to work on it, I'd like to see some tests or a comment explaining why changing `json_error_invalid_syntax` to `json_error_premature_end_of_input` is correct without more logic. Is it so because `saved_text` is empty only when end of input has been reached?

akheron:
Actually, looking a bit closer, this is quite clear as it is. Merged, thanks!

PR #389, opened 2018-01-23
Atomic references and thread safe json_dump.

### LINKED ISSUE #387: Use atomic built-ins to change json->refcount.
I would like to make `json_decref` / `json_incref` thread safe.  This would allow use of a single JSON object from multiple threads as long as the object were not manipulated (no adding/removing object fields or array elements).  For my use Asterisk PBX will be able to reduce jansson memory usage by bumping the reference instead of cloning the object.

A couple questions:
* Are the existing configure.ac tests adequate or do we need to check for every __sync / __atomic function we use?
* Do we need to move json_decref / json_incref to value.c and lose inlining or add results of the __sync / __atomic tests to jansson_config.h?  Keeping it in the header would create a requirement that programs using jansson use a compiler that supports the atomic feature specified by jansson_config.h.

coreyfarrell:
Fixes #387.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,8 +303,8 @@ else()
    set (JSON_INLINE)
 endif()
 
-check_c_source_compiles ("int main() { unsigned long val; __sync_bool_compare_and_swap(&val, 0, 1); return 0; } " HAVE_SYNC_BUILTINS)
-check_c_source_compiles ("int main() { char l; unsigned long v; __atomic_test_and_set(&l, __ATOMIC_RELAXED); __atomic_store_n(&v, 1, __ATOMIC_RELEASE); __atomic_load_n(&v, __ATOMIC_ACQUIRE); return 0; }" HAVE_ATOMIC_BUILTINS)
+check_c_source_compiles ("int main() { unsigned long val; __sync_bool_compare_and_swap(&val, 0, 1); __sync_add_and_fetch(&val, 1); __sync_sub_and_fetch(&val, 1); return 0; } " HAVE_SYNC_BUILTINS)
+check_c_source_compiles ("int main() { char l; unsigned long v; __atomic_test_and_set(&l, __ATOMIC_RELAXED); __atomic_store_n(&v, 1, __ATOMIC_RELEASE); __atomic_load_n(&v, __ATOMIC_ACQUIRE); __atomic_add_fetch(&v, 1, __ATOMIC_ACQUIRE); __atomic_sub_fetch(&v, 1, __ATOMIC_RELEASE); return 0; }" HAVE_ATOMIC_BUILTINS)
 
 set (JANSSON_INITIAL_HASHTABLE_ORDER 3 CACHE STRING "Number of buckets new object hashtables contain is 2 raised to this power. The default is 3, so empty hashtables contain 2^3 = 8 buckets.")
 
--- a/configure.ac
+++ b/configure.ac
@@ -38,25 +38,33 @@ AC_CHECK_FUNCS([close getpid gettimeofday localeconv open read sched_yield strto
 AC_MSG_CHECKING([for gcc __sync builtins])
 have_sync_builtins=no
 AC_TRY_LINK(
-  [], [unsigned long val; __sync_bool_compare_and_swap(&val, 0, 1);],
+  [], [unsigned long val; __sync_bool_compare_and_swap(&val, 0, 1); __sync_add_and_fetch(&val, 1); __sync_sub_and_fetch(&val, 1);],
   [have_sync_builtins=yes],
 )
 if test "x$have_sync_builtins" = "xyes"; then
   AC_DEFINE([HAVE_SYNC_BUILTINS], [1],
     [Define to 1 if gcc's __sync builtins are available])
+  json_have_sync_builtins=1
+else
+  json_have_sync_builtins=0
 fi
+AC_SUBST([json_have_sync_builtins])
 AC_MSG_RESULT([$have_sync_builtins])
 
 AC_MSG_CHECKING([for gcc __atomic builtins])
 have_atomic_builtins=no
 AC_TRY_LINK(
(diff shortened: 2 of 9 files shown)


coreyfarrell:
One comment I don't really like the `json_unsafe_ref` symbol name but don't have a better idea.

akheron:
Thanks for the PR!

What's the use case of `json_set_add_fetch_func`/`json_get_add_fetch_func`? Their downside seems to be that `json_incref` and `json_decref` need a function call to change the refcount on systems where __sync/__atomic are not supported.

What I was suggesting (in the comment that was unclear to you) was that in `jansson.h` you could just `#ifdef` on some constant that denotes whether the __sync/__atomic support is available. If it's not, just use plain old `++json->refcount`/`--json->refcount`.

Users could also `#ifdef` on the same constant in their own code to see whether refcount manipulation is thread-safe or not.

coreyfarrell:
In the case where __sync/__atomic are not available we (Asterisk) would use json_set_add_fetch_func provide a function which would use a pthread mutex to prevent multiple threads from manipulating references at the same time.  I assume that you don't want jansson linking to pthread so that's why I went with this approach.  In asterisk it is our fall back of last resort to use a global mutex to protect reference counting.

ploxiln:
As the alternative to atomic reference counting, you could avoid sharing json objects between threads, instead duplicating them for another thread.

coreyfarrell:
I just used grep to check every `*.c` in Asterisk.  2972 lines in 105 sources match `ast_json_` (the namespace our wrapper functions use).  Auditing that code for situations where duplication is required would be impractical.

We currently do use mutex when our own calls `json_decref` or `json_incref`, but our wrapper is not run by libjansson itself.  It's not always easy to predict when this could happen in large multithreaded applications, thus why this current PR gives a way to intercept those unsafe references.

I've opened a thread in the Asterisk developers mailing list about having the next version simply drop support for all compilers which do not have __sync or __atomic.  I'd like to put this PR on hold for a few days until I get responses (or not) from other Asterisk developers.

akheron:
> I've opened a thread in the Asterisk developers mailing list about having the next version simply drop support for all compilers which do not have __sync or __atomic.  I'd like to put this PR on hold for a few days until I get responses (or not) from other Asterisk developers.

All right, let's wait for their input.

coreyfarrell:
Thanks for being patient.  As of this morning the master branch of Asterisk will no longer compile unless __atomic or __sync operators are available.  I've updated this PR to only provide thread safe reference counting if supported by the compiler.

I made some tweaks to the reference counting code since my first commit.  Now I'm using __ATOMIC_ACQUIRE for json_incref and __ATOMIC_RELEASE for json_decref.  This tells the compiler and CPU that it's not allowed to hoist code above json_incref and it's not allowed to hoist json_decref.  This is unlikely to be any different in practice but should help to protect users of jansson against inappropriate code reorders.

akheron:
Thanks, looks a lot better now!

I'm a bit worried about the new circular reference check. It's nice that the `visited` field was removed from arrays and objects, but the loop in `dumper_init()` makes the encoder O(n²) (n being the depth of the value tree). Encoding might become very slow in a pathological case where there's a deep chain of arrays/objects.

Use a hashtable to simulate a set which stores seen `json_t` pointers. When starting to dump an array/object, check that the current `json_t` pointer is not yet in the set and add the pointer to the set. Remove the pointer from the set when dumping the pointer's value is done. 

A similar approach is used in `json_unpack()` to check that all object keys are unpacked, see https://github.com/akheron/jansson/blob/9a1d9c88fce268b812453d0d23237ee805b9bc34/src/pack_unpack.c#L421.

The downside is that this might involve quite a bit of work, as the current hashtable implementation only supports mappings with string keys. Also, it might feel like using a sledgehammer to crack a nut...

Another option is to leave the `visited` flag there, and fix the parallel encoding problem later. Or does Asterisk depend on encoding shared objects in multiple threads concurrently?

coreyfarrell:
We use a global mutex to ensure only one thread can encode at a time.  I'll use the sledgehammer, the goal of this is for all read-only operations to be safe without locking.  Our current approach is completely thread safe but it could also be a nasty bottleneck.  Especially with the JANSSON_THREAD_SAFE being defined, I want that to be true.

I'll take a look at this tomorrow when I'm on a computer.
PS when you get a chance please tag me on any issues you view as blockers for 2.11.  I'll do what I can to address blockers as I'm very interested in seeing a thread safe release.

coreyfarrell:
I've reworked the dump.c patch to use hashtable_t (without modifying hashtable.c).  About the cost of snprintf, I created the following test case:
```c
#include <stdio.h>
#include <inttypes.h>

int main(int argc, char *argv[])
{
	intptr_t i;
	void *myptr;
	char key[2 + (sizeof(myptr) * 2) + 1];

	for (i = 0; i < 1024*1024; i++) {
		myptr = (void*)i;
		snprintf(key, sizeof(key), "%p", myptr);
	}
	printf("%s\n", key);
	return 0;
}
```
I then ran `gcc mytest.c -o mytest && time ./mytest`.  This reported 0.1 seconds real and user time.  If someone is dumping something with 1 million objects/arrays the 0.1 seconds will be trivial.

akheron:
Thanks, and sorry for the delay.

It's almost perfect now. I took a liberty to clarify the documentation around thread-safety a bit, and rename `JANSSON_THREAD_SAFE` to `JANSSON_THREAD_SAFE_REFCOUNT`. The renaming will help if Jansson will somehow support full thread-safety in the future :)

My changes are in this commit: https://github.com/akheron/jansson/commit/f44921e1766b4584bb239593ce7e7ea3242c99c3. What do you think?

coreyfarrell:
Your changes look good, I pulled it back to my branch (though I didn't expect it to create a merge commit).  I don't have a problem with defining JANSSON_THREAD_SAFE_REFCOUNT to 1, as long as we never define it to 0.  Defining to 0 would only serve to make feature detection more difficult.

akheron:
Merged, thanks!

PR #393, opened 2018-02-08
Add json_sprintf and json_vsprintf

### LINKED ISSUE #392: Add json_sprintf
Hi Petri!
I use intensely your great library. There is several sentence that I repeat a lot: convert a number to string and the put it in an object. Something like this:

    char sevent[32];
    snprintf(sevent, sizeof(sevent), "%d", event);
    json_object_set_new(jn_gps_msg, "MessageID", json_string(sevent));

Why not a shortcut like this?:

    json_object_set_new(jn_gps_msg, "MessageID", json_sprintf("%d", event));

being the return of 

    json_t * json_sprintf(const char *format, ...)

 a json string.

Thanks for listening me.
Niyamaka

akheron:
Fixes #392

### DIFF
--- a/src/value.c
+++ b/src/value.c
@@ -787,6 +787,40 @@ static json_t *json_string_copy(const json_t *string)
     return json_stringn_nocheck(s->value, s->length);
 }
 
+json_t *json_vsprintf(const char *fmt, va_list ap) {
+    int length;
+    char *buf;
+    va_list aq;
+    va_copy(aq, ap);
+
+    length = vsnprintf(NULL, 0, fmt, ap);
+    if (length == 0)
+        return json_string("");
+
+    buf = jsonp_malloc(length + 1);
+    if (!buf)
+        return NULL;
+
+    vsnprintf(buf, length + 1, fmt, aq);
+    if (!utf8_check_string(buf, length)) {
+        jsonp_free(buf);
+        return NULL;
+    }
+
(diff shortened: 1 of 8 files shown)


src/value.c
```diff
@@ -787,6 +787,40 @@ static json_t *json_string_copy(const json_t *string)
...
+    length = vsnprintf(NULL, 0, fmt, ap);
+    if (length == 0)
+        return json_string("");
+
+    buf = jsonp_malloc(length + 1);
+    if (!buf)
+        return NULL;
+
+    vsnprintf(buf, length + 1, fmt, aq);
+    result = json_stringn(buf, length);
```
coreyfarrell: An extra allocation/string copy could be avoided if we directly called `utf8_check_string` then use `jsonp_stringn_nocheck_own`.

Your call if you want to do take this suggestion or ignore it.

src/value.c
```diff
@@ -787,6 +787,40 @@ static json_t *json_string_copy(const json_t *string)
     return json_stringn_nocheck(s->value, s->length);
 }
 
+static json_t *json_vsprintf(const char *fmt, va_list ap) {
```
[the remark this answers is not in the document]
akheron: Done

src/value.c
akheron: Fixed, good catch!

akheron:
Merged, thanks @coreyfarrell for reviewing!

PR #397, opened 2018-02-12
add gcc attribute for json_sprintf and json_vsprintf

Mephistophiles:
Added gcc attribute for strict validation format string.

### DIFF
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -348,6 +348,33 @@ static json_t *pack_string(scanner_t *s, va_list *ap)
     }
 }
 
+static json_t *pack_object_inter(scanner_t *s, va_list *ap, int need_incref)
+{
+    int nullable;
+    json_t *json;
+
+    next_token(s);
+    nullable = token(s) == '?';
+    if (!nullable)
+        prev_token(s);
+
+    json = va_arg(*ap, json_t *);
+
+    if (json) {
+        return need_incref ? json_incref(json) : json;
+    }
+    if (nullable) {
+        return json_null();
+    } else if (s->next_token.token == '*') {
+        /* FIXME: need a more elegant way. */
+        return NULL;
+    }
+
+    set_error(s, "<args>", json_error_null_value, "NULL object key");
+    s->has_error = 1;
+    return NULL;
+}
+
 static json_t *pack(scanner_t *s, va_list *ap)
 {
     switch(token(s)) {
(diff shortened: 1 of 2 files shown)


coreyfarrell:
This is a good idea but I believe this attribute is supported by more than just gcc (for example clang/llvm).  I'm not sure if we should use the attribute for `#if defined(__GNUC__) || defined(__clang__)` or if we should add a check to configure.ac and CMakeLists.txt, then add `#define attribute_printf(fmt, args) @ATTRIBUTE_PRINTF@` to src/jansson_config.h.in.

@akheron what do you think?

akheron:
In my opinion, the check for `__GNUC__` or `__clang__` is the way to go, no need to involve configure.ac or CMakeLists.txt.

src/jansson.h
```diff
@@ -291,7 +291,14 @@ int json_vunpack_ex(json_t *root, json_error_t *error, size_t flags, const char
 
 /* sprintf */
 
+#ifdef __GNUC__
+__attribute__((format(printf, 1, 2)))
+#endif
```
akheron: Is this attribute correct before the function declaration? In the examples in https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html, attributes are always specified after the function declataion.

coreyfarrell:
> In my opinion, the check for __GNUC__ or __clang__ is the way to go, no need to involve configure.ac or CMakeLists.txt.

Any opinion on creating a block that defines `JSON_ATTR_PRINTF(fmt, first_arg_idx)` to produce the attribute or nothing depending on the compiler?  This may seem like overkill for the 2 uses of this attribute type but other attributes would be useful to enhance the compiler warnings, for example `warn_unused_result` on `strbuffer_init` would get the compiler to warn us when ignore that allocation failure.  I think having the check for `__GNUC__` or `__clang__` on every function that uses an attribute would make the header more difficult to read.

akheron:
> Any opinion on creating a block that defines JSON_ATTR_PRINTF(fmt, first_arg_idx) to produce the attribute or nothing depending on the compiler?

Go for it. Let's see how it looks :)

src/jansson.h
Mephistophiles: Yes. You can use the attributes before and after declaring the function. If you want - I can move the attribute after the function declaration.

Mephistophiles:
It was offered a more elegant solution, so I close this pull req.

PR #398, opened 2018-02-15
Improve test coverage.

coreyfarrell:
* Test equality of different length strings.
* Add tab to json_pack whitespace test.
* Test json_sprintf with empty result and invalid UTF.
* Test json_get_alloc_funcs with NULL arguments.
* Test invalid arguments.
* Add test_chaos to test allocation failure code paths.
* Remove redundant json_is_string checks from json_string_equal and
  json_string_copy.  Both functions are static and can only be called
  with a json string.

Fixes to issues found by test_chaos:
* Fix crash on OOM in pack_unpack.c:read_string().
* Unconditionally free string in string_create upon allocation failure.
  Update load.c:parse_value() to reflect this.  This resolves a leak on
  allocation failure for pack_unpack.c:pack_string() and
  value.c:json_sprintf().

Although not visible from CodeCoverage these changes significantly
increase branch coverage.  Especially in src/value.c where we previously
covered 67.4% of branches and now cover 96.3% of branches.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -487,6 +487,7 @@ if (NOT JANSSON_WITHOUT_TESTS)
    set(api_tests
          test_array
          test_copy
+         test_chaos
          test_dump
          test_dump_callback
          test_equal
--- a/src/load.c
+++ b/src/load.c
@@ -829,10 +829,8 @@ static json_t *parse_value(lex_t *lex, size_t flags, json_error_t *error)
             }
 
             json = jsonp_stringn_nocheck_own(value, len);
-            if(json) {
-                lex->value.string.val = NULL;
-                lex->value.string.len = 0;
-            }
+            lex->value.string.val = NULL;
+            lex->value.string.len = 0;
             break;
         }
 
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -159,7 +159,10 @@ static char *read_string(scanner_t *s, va_list *ap,
         return (char *)str;
     }
 
-    strbuffer_init(&strbuff);
+    if(strbuffer_init(&strbuff)) {
+        set_error(s, "<internal>", json_error_out_of_memory, "Out of memory");
+        s->has_error = 1;
+    }
 
     while(1) {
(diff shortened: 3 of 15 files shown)


coreyfarrell:
Stuff that still needs coverage but will not be added to this PR:
* json_load_callback - no tests call this function
* json_unpack - need to test allocation failures and bad arguments.

I'm not sure why I couldn't get snprintf to work on the older VS compilers, I do not plan to investigate this any further.

These changes were tested with `make VALGRIND=1 distcheck`.

coreyfarrell:
Additional comment, in my own test I modified cmake/CodeCoverage.cmake.  Inside `SETUP_TARGET_FOR_COVERAGE` I added `--rc lcov_branch_coverage=1` to both `lcov` calls and `--branch-coverage` to `genhtml`.  I'm not sure if you'd be open to adding branch coverage by default or if you know of a way to make it optional.  I'm not that familiar with cmake to know how to make it optional.

akheron:
Hi and thanks for the great PR! Enabling branch coverage by default would be great. Would it then also be reported by coveralls?

akheron:
The chaos test is awesome :)

coreyfarrell:
I submitted #400 to enable branch coverage, it doesn't cause a problem for coveralls but they do not appear to support reporting branch coverage.  I'd still like it to be merged if it's OK with you as it'll be useful for running tests locally and hunting for uncovered code / conditions (a lot easier than keeping the change in `git stash` forever).

Regarding #364 I have seen that PR.  test_chaos picked up the `strbuffer_init` failure in `read_string`.  I don't disagree with you about adding more coverage for strbuffer failures but this PR is already pretty large, I'd rather keep the scope where it is and use separate PR's for further expansion.  I do not want to address the other bugs shown in #364 without tests to cover those failures - the handling of these error branches in pack_unpack.c are trickier since #374.  For now test_chaos mostly targets src/value.c and the `json_pack` parts of pack_unpack.c as these are the API's I'm most familiar with.  I did not include redundant direct testing of src/value.c as json_pack covers basic allocation functions like `json_object()`.

PR #404, opened 2018-03-06
Build: Add JANSSON_ATTRS macro.

coreyfarrell:
This macro is used to conditionally generate GCC/CLANG __attribute__
declarations if supported.

This allows the compiler to produce warnings on certain incorrect
usages.  json_sprintf and json_vsprintf will produce warnings on invalid
format string.  Many functions will produce a warning if the result is
unused.  Specifically functions which allocate new objects will warn if
the result is ignored as this always results in a memory leak.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -493,6 +493,7 @@ if (NOT JANSSON_WITHOUT_TESTS)
          test_equal
          test_load
          test_loadb
+         test_load_callback
          test_number
          test_object
          test_pack
--- a/src/hashtable.h
+++ b/src/hashtable.h
@@ -55,7 +55,7 @@ typedef struct hashtable {
  *
  * Returns 0 on success, -1 on error (out of memory).
  */
-int hashtable_init(hashtable_t *hashtable);
+int hashtable_init(hashtable_t *hashtable) JANSSON_ATTRS(warn_unused_result);
 
 /**
  * hashtable_close - Release all resources used by a hashtable object
--- a/src/jansson.h
+++ b/src/jansson.h
@@ -39,6 +39,12 @@ extern "C" {
 #define JANSSON_THREAD_SAFE_REFCOUNT 1
 #endif
 
+#if defined(__GNUC__) || defined(__clang__)
+#define JANSSON_ATTRS(...) __attribute__((__VA_ARGS__))
+#else
+#define JANSSON_ATTRS(...)
+#endif
+
 /* types */
 
 typedef enum {
@@ -185,7 +191,7 @@ static JSON_INLINE enum json_error_code json_error_code(const json_error_t *e) {
 
 void json_object_seed(size_t seed);
 size_t json_object_size(const json_t *object);
-json_t *json_object_get(const json_t *object, const char *key);
+json_t *json_object_get(const json_t *object, const char *key) JANSSON_ATTRS(warn_unused_result);
 int json_object_set_new(json_t *object, const char *key, json_t *value);
 int json_object_set_new_nocheck(json_t *object, const char *key, json_t *value);
 int json_object_del(json_t *object, const char *key);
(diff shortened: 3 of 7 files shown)


coreyfarrell:
Technically `strbuffer_init` should have `warn_unused_result` but this actually produces 2 warnings currently.  I haven't had a chance to deal with those warnings yet but I wanted to post this so it can be reviewed in comparison to #397.  The plan is to post a follow-up that will deal with the 2 incorrect callers of `strbuffer_init` and add the `warn_unused_result` attribute at the same time.

coreyfarrell:
OK I had a little extra time so I've addressed the callers of strbuffer_init and added the attribute.

Additional note: I'm unsure what's going on with coveralls.   I don't think this PR should have broken it, plus I would expect it to post an error if that were the case.  I'm guessing something about the integration between coveralls and travis-ci is currently offline?

akheron:
Good work!

JamesAllwrightAA:
The JANSSON_ATTRS() macro relied on variadic macro support, which is a C99
feature. I am seeing failures where the wrong number of parameters is reported.
As a proposed fix for this, you could use
#if defined(__GNUC__) || defined(__clang__)
json_t *json_sprintf(const char *fmt, ...) JANSSON_ATTRS(warn_unused_result, format(printf, 1, 2));
json_t *json_vsprintf(const char *fmt, va_list ap) JANSSON_ATTRS(warn_unused_result, format(printf, 1, 0));
#else
json_t *json_sprintf(const char *fmt, ...);
json_t *json_vsprintf(const char *fmt, va_list ap);
#endif

ploxiln:
please use fenced code blocks to make your code examples readable, like so:

    ```
    your code pasted here
    with the surrounding triple-backticks on their own separate lines
    ```

So right now we have this:

```c
#if defined(__GNUC__) || defined(__clang__)
#define JANSSON_ATTRS(...) __attribute__((__VA_ARGS__))
#else
#define JANSSON_ATTRS(...)
#endif
```

and apparently your compiler can't handle the:

```c
#else
#define JANSSON_ATTRS(...)
```

but your proposal would still present a variadic macro to your compiler:

```c
#else
json_t *json_sprintf(const char *fmt, ...);
```

so that wouldn't work either.

JamesAllwrightAA:
Just to clarify the previous comment : 

The problem is C preprocessor (CPP) handling of variadic macros, not C 
compiler handling of variadic functions. I have tested my proposed fix and 
it does resolve the problem on my system. I agree that there are still variadic 
C functions in the code, but these are OK because they are supported in an 
earlier C standard (C89) whereas CPP variadic macros are not supported
until C99.

JamesAllwrightAA:
Here is another suggestion for how the code could be re-written to avoid
variadic macros. It involves more code changes, but arguably is more
readable because there is less magic going on in the macro.
```
#if defined(__GNUC__) || defined(__clang__)
#define JANSSON_ATTRS(x) x
#else
#define JANSSON_ATTRS(x)
#endif
```
we then put
```
json_t *json_object_get(const json_t *object, const char *key) JANSSON_ATTRS(__attribute__((warn_unused_result)));
```
and likewise in the other places where we use this construct
and around line 300
```
json_t *json_sprintf(const char *fmt, ...) JANSSON_ATTRS(__attribute__((warn_unused_result, format(printf, 1, 2))));
json_t *json_vsprintf(const char *fmt, va_list ap) JANSSON_ATTRS(__attribute__((warn_unused_result, format(printf, 1, 0))));
```

ploxiln:
> The problem is C preprocessor (CPP) handling of variadic macros, not C compiler handling of variadic functions.

Ah, you're right, my mistake.

I do think the direction of your second suggestion is better. To tweak it a bit further, maybe something like:

```c
#if defined(__GNUC__) || defined(__clang__)
#define JANSSON_ATTRS(x) __attribute(x)
#else
#define JANSSON_ATTRS(x)
#endif

json_t *json_sprintf(const char *fmt, ...) JANSSON_ATTRS((warn_unused_result, format(printf, 1, 2)));
```

coreyfarrell:
@JamesAllwrightAA / @ploxiln the suggestion of altering JANSSON_ATTRS to accept a single argument is much better than using conditional code for every function declaration that uses attributes.  Would you like to submit a PR for this?  Sorry I just got back from being AFK for nearly 2 weeks, I'm just buried at the moment so I won't have time to pursue this right now.  If you can't post a PR please open a new issue linked to this PR so I have something open to remind me later.

PR #408, opened 2018-03-11
CMake improvements, make package relocatable

isaachier:
The current build does not allow one to move the library to another directory. By writing a config with absolute paths, the dependencies will incorrectly check the build directory instead of the current location. This patch uses relative paths from the CMake config to keep the config valid even if the entire package is moved.


akheron:
Hi and thanks for the PR.

I've felt that the CMake configuration needs cleanup for a long time now. The problem is that my knowledge about CMake is close to zero, and I don't really mind learning it to be honest.

Do you think this PR cleans as much of the CMake config as possible? I think at least most of the comment near the top of `CMakeLists.txt` should also be removed. Is there anything else after that?

akheron:
@JoakimSoderberg do you happen to have time to comment on this PR?

isaachier:
I deleted the original configs. I added a new one in `cmake/Config.cmake.in`. I've love to help enhance the CMake build in general. These changes were minimal to ensure compliance with the Hunter package manager (https://github.com/ruslo/hunter). My project relies on Hunter to download and build a relocatable Jansson.

CMakeLists.txt
```diff
@@ -608,80 +607,67 @@ set(VERSION     ${JANSSON_DISPLAY_VERSION})
...
+
+  # Install the configs.
+  install(FILES
+      ${CMAKE_CURRENT_BINARY_DIR}/cmake/janssonConfig.cmake
+      ${CMAKE_CURRENT_BINARY_DIR}/cmake/janssonConfigVersion.cmake
+      DESTINATION "${JANSSON_INSTALL_CMAKE_DIR}")
+
+  # Install exports for the install-tree.
+  install(EXPORT janssonTargets
+          NAMESPACE jansson::
```
JoakimSoderberg: I'm new to this.

>  The NAMESPACE option will prepend <namespace> to the target names as they are written to the import file. By default the generated file will be called <export-name>.cmake

https://cmake.org/cmake/help/v3.0/command/install.html?highlight=install

But as I understand this is not the same thing as **C++ namespace**. Only used for not colliding with other CMake projects.
https://stackoverflow.com/questions/25983741/cmake-enforcing-a-relationship-between-namespaces-and-library-file-names
> Also note the mechanism is only concerned about avoiding name clashes in CMake scripts, not about name clashes on the filesystem. The latter need to be resolved on another level.

isaachier:
Definitely not the same as C++ namespace. In general, cmake doesn't do symbol renaming or anything libtool might do. This pull request is a bit outdated so let me update it before merging.

akheron:
Merged, thanks!

PR #409, opened 2018-03-12
Fix error handling in json_pack

Mephistophiles:
Fixed a bug where the error message was not filled if an empty object
was passed to the json_pack.

### DIFF
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -348,6 +348,36 @@ static json_t *pack_string(scanner_t *s, va_list *ap)
     }
 }
 
+static json_t *pack_object_inter(scanner_t *s, va_list *ap, int need_incref)
+{
+    json_t *json;
+    char ntoken;
+
+    next_token(s);
+    ntoken = token(s);
+
+    if (ntoken != '?')
+        prev_token(s);
+
+    json = va_arg(*ap, json_t *);
+
+    if (json)
+        return need_incref ? json_incref(json) : json;
+
+    switch (ntoken) {
+        case '?':
+            return json_null();
+        case '*':
+            return NULL;
+        default:
+            break;
+    }
+
+    set_error(s, "<args>", json_error_null_value, "NULL object key");
+    s->has_error = 1;
+    return NULL;
+}
+
 static json_t *pack(scanner_t *s, va_list *ap)
 {
     switch(token(s)) {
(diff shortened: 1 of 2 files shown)


src/pack_unpack.c
```diff
@@ -348,6 +348,33 @@ static json_t *pack_string(scanner_t *s, va_list *ap)
     }
 }
 
+static json_t *pack_object_inter(scanner_t *s, va_list *ap, int need_incref)
+{
+    int nullable;
+    json_t *json;
+
+    next_token(s);
+    nullable = token(s) == '?';
```
coreyfarrell: Instead of `int nullable` could you create a var `char ntoken` and set it to `token(s)`?  This way you can check for `ntoken == '?'` or `ntoken == '*'` below.  I'd prefer this over reading `s->next_token.token` directly.  I know other code reads `s->next_token.token` directly but I feel we should move away from this as it's only set by calls to `prev_token()`, not calls to `next_token()`.

src/pack_unpack.c
```diff
@@ -348,6 +348,33 @@ static json_t *pack_string(scanner_t *s, va_list *ap)
...
+    if (!nullable)
+        prev_token(s);
+
+    json = va_arg(*ap, json_t *);
+
+    if (json) {
```
coreyfarrell: The rest of the code does not use `{}` for single statements conditional branches, we should do the same in this function to be consistent.

src/pack_unpack.c
```diff
@@ -348,6 +348,33 @@ static json_t *pack_string(scanner_t *s, va_list *ap)
...
+        return need_incref ? json_incref(json) : json;
+    }
+    if (nullable) {
+        return json_null();
+    } else if (s->next_token.token == '*') {
```
coreyfarrell: The previous branch return's you could drop the `else` and just start a new branch.  IMO this makes the code easier to read.

src/pack_unpack.c
Mephistophiles: Fixed. Thanks!

src/pack_unpack.c
Mephistophiles: Fixed. Thanks!

src/pack_unpack.c
Mephistophiles: I rewrote to switch, look, that's better?

coreyfarrell:
@Mephistophiles one more thing, can you add a line `Fixes #271` to the commit message so github closes that bug when this is merged?

Mephistophiles:
@coreyfarrell done!

src/pack_unpack.c
```diff
@@ -348,6 +348,37 @@ static json_t *pack_string(scanner_t *s, va_list *ap)
...
+
+    if (json)
+        return need_incref ? json_incref(json) : json;
+
+    switch (ntoken) {
+        case '?':
+            return json_null();
+        case '*':
+            /* FIXME: need a more elegant way. */
+            return NULL;
```
akheron: What makes this not elegant?

akheron: In other words: Why do you think this is not elegant? :)

akheron:
Merged, thanks!

PR #412, opened 2018-03-21
Fix JSON_HAVE_SYNC_BUILTINS define usage

Mephistophiles:
Fixed warning on the build:
/usr/include/jansson.h:38:5: warning: "JSON_HAVE_ATOMIC_BUILTINS" is not defined [-Wundef]

For fix this warning must be use 'defined' keywords.


akheron:
Merged, thanks!

PR #415, opened 2018-04-28
CMake variable to number comparison changed to provide meaningful error

palmerc:
In the code using check_size_type a value is assigned to a variable like INT. Then the value is checked in the if-block if it is equal to a value like 4.

`if (HAVE_INT AND ("${INT}" EQUAL 4))`

This results in an error if check_size_type fails because the value of INT will be "" resulting in an error in the arguments to the if-block. A better comparison is

`if (HAVE_INT AND (INT EQUAL 4))`

This will fall through as intended to the `message( FATAL_ERROR "Message" )` section.


akheron:
Hi and thanks for the PR!

- Could you separate the whitespace-only changes to a separate commit to make it easier to see what actually changed.
- I see the `-fPIC` handling was also changed and the `POSITION_INDEPENDENT_CODE` property is now set. Was this intentional or necessary? According to CMake documentation, `POSITION_INDEPENDENT_CODE` is set by default for `SHARED` library targets.

akheron:
I don't see the point of applying PIC to the static library. @JoakimSoderberg do you happen to remember if this is intentional? It seems to me that the whole `CMakeLists.txt` could use some major cleaning up if anyone knowing CMake has time to do it.

palmerc:
If you include the static library in a shared library.

akheron:
I merged #408 and this PR now conflicts. If you think this is still relevant, could you please rebase it?

palmerc:
@akheron I believe you'll find this rebased

akheron:
Merged, thanks!

PR #423, opened 2018-07-14
Deal with warnings under gcc 8.

coreyfarrell:
Recent versions of gcc have introduced compiler warnings for string
operations that could be truncated.  This caused problems with -Werror.
src/error.c used strncpy to write "..." to a string, but skipped writing
the NUL terminator.  Switch this to use memcpy.  src/load.c produced
warnings from snprintf writing error strings that could be truncated.
Added code to autotools build to detect `-Wno-format-truncation', add it
to AM_CFLAGS if supported.


akheron:
Merged, thanks! And sorry for the delay.

I wonder if there's a way to check for `-Wno-format-truncation` support in CMake also.

SamyDgz (approved):

SamyDgz (approved):

Issue #10, opened 2010-12-20
Add a decoding flag to get all numbers as strings


akheron:
All numbers or only overflowing numbers, or maybe two separate flags?

keenerd:
For an example of how another library does this:  yajl-tree simply stores the original string and lets you access it directly: https://github.com/lloyd/yajl/blob/master/src/api/yajl_tree.h#L81

akheron:
Yeah, this is also an option. But allocating the extra memory to store input strings of every number doesn't sound good for people using this on embedded devices. It could of course be enabled by a decoding flag to make it optional.

Would if be beneficial to do this also for strings? They have many possible input forms because any Unicode code point can be escaped with `\u` or represented directly in UTF-8.

dbelliveau:
Because of the decimal precision issue we made a change to our copy of Jansson to allow all numbers to be treated as strings. This might not be efficient for many use cases but for ours it's actually more efficient. In brief:

* New flag JSON_DECODE_NUMBER_AS_STRING causes json_load* to decode all numbers as strings.
* New int field in json_string_t to indicate if a string is a number. Set to true when decoding if the source field is a number. When encoding, if true json_dump* will encode the string as a number (no quotes).
* New function json_string_is_number() returns true if a string is number.
* New function json_string_set_is_number() to set the "string is a number" indicator.

If there's interest in adding this to Jansson let me know and I'll create a pull request.

nertpinx:
@dbelliveau If it's still possible to submit your patches as PR, then it might help with #425 (my guess is that it's taking similar approach but it comes from a different source).

coreyfarrell:
@akheron any reason you are against creating a new `json_type` for this (maybe `JSON_RAW`)?  My complaint about the current patches is that it expands storage of json_string_t.  I realize adding a new type will require more care but I think it's worth the effort.  One thing I noticed about the patch @keenerd created is that JSON_SNUMBER is added to the middle of the enum, it would need to be last on the enum otherwise ABI would change.  Naming it JSON_RAW instead of JSON_SNUMBER would leave us the option to use that same type to support raw / non-decoded strings as well (possible micro-optimization for some use cases).

akheron:
I'm not against adding a new type if it's not only for working around quirks of floating point formatting.

What are the use cases that we're trying to support here? I think I've lost track :)

Supporting numbers that don't have a native representation ("bignums") come to my mind, but are there other use cases?

akheron:
[...]

PR #425, opened 2018-07-16
Add JSON_DECODE_NUMBER_AS_STRING flag

montegoulding:
This patch adds a flag to decode all numbers as strings to allow for JSON containing
numbers that overflow either int or real and/or in cases where a higher degree of
precision is required and string to number conversion can be done outside Jansson.


src/jansson.h
```diff
@@ -319,6 +319,7 @@ json_t *json_deep_copy(const json_t *value) JANSSON_ATTRS(warn_unused_result);
 #define JSON_DECODE_ANY         0x4
 #define JSON_DECODE_INT_AS_REAL 0x8
 #define JSON_ALLOW_NUL          0x10
+#define JSON_DECODE_NUMBER_AS_STRING          0x12
```
akheron: The value should be `0x20` to not have same bits with other flags.

akheron:
Hi and thanks for the PR!

Are you sure this is the right approach? I'm just thinking about how the user is able to know which strings are actually numbers.

src/jansson.h
montegoulding: woops... fixed

montegoulding:
Hi @akheron.

We are dealing with a JSON files that may have _very_ large numbers and we are hitting both overflow and precision issues. We are also using the values in a system where everything is a string until it _must_ be treated as a number so we actually have to convert back into string anyway. It may be that our use case is too much an edge case for you though. Probably it would be nicer to have all numbers be a reference to the string and then an api to get as type so you know it's a number... but that would be a much bigger patch ;-)

I also thought about adding  `JSON_OVERFLOW_NUMBER_AS_STRING` but all numbers as strings is better for our situation.

akheron:
There's an old issue #10 about this. Especially this comment https://github.com/akheron/jansson/issues/10#issuecomment-257018475 is interesting. The author has added a flag to strings to denote whether the string contains a number.

montegoulding:
Aha... a PR for that would have saved me some time ;-)

PR #433, opened 2018-08-28
no-float no-errno no-file

pkuzco:
- Added an option to remove the dependency on errno
- Allow building JANSSON without real-numbers support
- Added an option to remove the dependency on the standard FILE, and let the user re-implement json_[load/dump]_file if they need to.


akheron:
I'm quite reluctant to merge these changes. I believe this would make maintaining the library much harder to keep the support for such a constrained environment.

Have you explored alternative JSON libraries that are specifically targeted for embedded devices?

pkuzco:
I have, but none of the other libraries provides a good api to work with.

Will close this PR, as I agree, these changes will likely complicate maintaining the code - but I will probably continue to maintain a separate fork that is more targetted to embedded platforms.

PR #436, opened 2018-09-03
removed various warnings

edgale:
after doing https://github.com/akheron/jansson/pull/435 I enabled a few more compiler warning flags with my configuration of the library. This is the stuff it came up with.
I tried to solve it as best as I can. But maybe some changes are not necessary. I just thought I share it if somethings of this is useful.

### DIFF
--- a/src/dump.c
+++ b/src/dump.c
@@ -351,7 +351,7 @@ static int do_dump(const json_t *json, size_t flags, int depth,
                 }
                 assert(i == size);
 
-                qsort((keys, size, sizeof(const char *), compare_keys);
+                qsort(keys, size, sizeof(const char *), compare_keys);
 
                 for(i = 0; i < size; i++)
                 {
--- a/src/dump.c
+++ b/src/dump.c
@@ -351,7 +351,7 @@ static int do_dump(const json_t *json, size_t flags, int depth,
                 }
                 assert(i == size);
 
-                qsort((void*)keys, size, sizeof(const char *), compare_keys);
+                qsort((keys, size, sizeof(const char *), compare_keys);
 
                 for(i = 0; i < size; i++)
                 {
@@ -366,7 +366,7 @@ static int do_dump(const json_t *json, size_t flags, int depth,
                     if(dump(separator, separator_length, data) ||
                        do_dump(value, flags, depth + 1, parents, dump, data))
                     {
-                        jsonp_free((void*)keys);
+                        jsonp_free(keys);
                         return -1;
                     }
 
@@ -375,7 +375,7 @@ static int do_dump(const json_t *json, size_t flags, int depth,
                         if(dump(",", 1, data) ||
                            dump_indent(flags, depth + 1, 1, dump, data))
                         {
-                            jsonp_free((void*)keys);
+                            jsonp_free(keys);
                             return -1;
                         }
                     }
@@ -383,7 +383,7 @@ static int do_dump(const json_t *json, size_t flags, int depth,
                     {
                         if(dump_indent(flags, depth, 0, dump, data))
                         {
-                            jsonp_free((void*)keys);
+                            jsonp_free(keys);
                             return -1;
                         }
                     }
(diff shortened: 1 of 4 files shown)


coreyfarrell:
Can you share info about your compiler and how your configured jansson (or if this is the result of a static analyzer)?  I tried:
`CFLAGS='-Wall -Wextra -Werror' ./configure && make` - this did not produce any warnings.

I don't have any complaint about initializing the variables to 0 (codepoint and seed) or removal of the unreachable 'return NULL' statements.  I'm not a fan of the explicit casts, I'm interested in knowing what asked you to make this changes.

edgale:
Hello @coreyfarrell ,
thanks for the responds.
I am using Visual Studio 17 v15.8.5 on Windows 10 with all warining enabled (/Wall) in the project settings.
For the configuration of jansson, in jansson_config.h 
<details>
<summary>jansson-config.h</summary>

```c
/*
 * Copyright (c) 2010-2016 Petri Lehtinen <petri@digip.org>
 *
 * Jansson is free software; you can redistribute it and/or modify
 * it under the terms of the MIT license. See LICENSE for details.
 *
 *
 * This file specifies a part of the site-specific configuration for
 * Jansson, namely those things that affect the public API in
 * jansson.h.
 *
 * The CMake system will generate the jansson_config.h file and
 * copy it to the build and install directories.
 */

#ifndef JANSSON_CONFIG_H
#define JANSSON_CONFIG_H

/* Define this so that we can disable scattered automake configuration in source files */
#ifndef JANSSON_USING_CMAKE
#define JANSSON_USING_CMAKE
#endif

/* Note: when using cmake, JSON_INTEGER_IS_LONG_LONG is not defined nor used,
 * as we will also check for __int64 etc types.
 * (the definition was used in the automake system) */

/* Bring in the cmake-detected defines */
#define HAVE_STDINT_H 1
/* #undef HAVE_INTTYPES_H */
#define HAVE_SYS_TYPES_H 1

/* Include our standard type header for the integer typedef */

#if defined(HAVE_STDINT_H)
#  include <stdint.h>
#elif defined(HAVE_INTTYPES_H)
#  include <inttypes.h>
#elif defined(HAVE_SYS_TYPES_H)
#  include <sys/types.h>
#endif


/* If your compiler supports the inline keyword in C, JSON_INLINE is
   defined to `inline', otherwise empty. In C++, the inline is always
   supported. */
#ifdef __cplusplus
#define JSON_INLINE inline
#else
#define JSON_INLINE inline
#endif


#define json_int_t long long
#define json_strtoint strtoll
#define JSON_INTEGER_FORMAT "I64d"


/* If locale.h and localeconv() are available, define to 1, otherwise to 0. */
#define JSON_HAVE_LOCALECONV 1

/* If __atomic builtins are available they will be used to manage
   reference counts of json_t. */
#define JSON_HAVE_ATOMIC_BUILTINS 0

/* If __atomic builtins are not available we try using __sync builtins
   to manage reference counts of json_t. */
#define JSON_HAVE_SYNC_BUILTINS 0

/* Maximum recursion depth for parsing JSON input.
   This limits the depth of e.g. array-within-array constructions. */
#define JSON_PARSER_MAX_DEPTH 2048

#endif

```
</details>


Concerning these cast:
https://github.com/akheron/jansson/blob/fc591913ac600c57a900a81c0b7e67a9b6e38bdc/src/dump.c#L354
VS17 gives the warning C4090: *funciton*: different const qualifier.

https://github.com/akheron/jansson/blob/fc591913ac600c57a900a81c0b7e67a9b6e38bdc/src/load.c#L177
VS17 gives the warning C4242: conversion from "int" to "char", possible data loss.

akheron:
I'm also not a fan of the explicit casts, especially the `(void *)` ones. Otherwise ok.

edgale:
Thank you for your feedback. I changed the explicit casts.

coreyfarrell:
Thanks for the updates.  I'm going to give it a couple days for others to potentially object before I merge.

SamyDgz (approved):

PR #438, opened 2018-09-24
json_pack: Improve handling of formats with '?' and '*'.

### LINKED ISSUE #437: json_pack "s?" error reporting issues.
This is mostly a reminder for myself to patch this when I get a chance.  In pack_unpack.c:read_string() single-string branch we set error "NULL string argument" on NULL string even if `?` is used.  This should not be an error.  Further we do not set `s->has_error = 1` for errors in the single-string branch which could cause invalid UTF-8 to be converted to json_null (for strings associated with `s?` format).

This code also needs to be checked for "s*", I think we need to skip setting "NULL string argument" error for that as well.

Ref https://gerrit.asterisk.org/#/c/asterisk/+/10194/1/main/json.c@619

coreyfarrell:
When NULL is received for an optional argument we should not set an
error message as this would block later error messages. If NULL is
received for a non-optional string we should set has_error. Set
has_error for UTF-8 errors to ensure optional strings with UTF-8
errors are not replaced with json_null(). Use 'purpose' argument in
NULL error messages of read_string.

Add error handling and tests for invalid formats where '+', '#', or '%'
is used on an optional string 's?' or 's\*'.

Fix NULL string error messages to use 'purpose'.

Refactor skipping of '\*' token, this is now handled by read_string and
pack_object_inter. This allows invalid format strings such as 's\*#' and
's\*+' to produce error messages.

Fixes #437

### DIFF
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -130,7 +130,7 @@ static json_t *pack(scanner_t *s, va_list *ap);
 /* ours will be set to 1 if jsonp_free() must be called for the result
    afterwards */
 static char *read_string(scanner_t *s, va_list *ap,
-                         const char *purpose, size_t *out_len, int *ours)
+                         const char *purpose, size_t *out_len, int *ours, int optional)
 {
     char t;
     strbuffer_t strbuff;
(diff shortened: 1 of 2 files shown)


coreyfarrell:
This will need some work on docs before it can be considered for merge.  I'm currently discussing the expanded 's?' functionality in the Asterisk dev channel.

coreyfarrell:
@akheron I'm ready for this to be considered for merge (assuming old versions of VS don't fail again). Changes were made from the initial patch after comments from @rmudgett9125 on the Asterisk developer channel.

akheron:
This is a very good set changes!

Do you think docs will still need to be updated?

coreyfarrell:
> Do you think docs will still need to be updated?

No, I was thinking that because an earlier version of the patch expanded functionality of 's?' and 's*'.  The current patches do not go against documented behavior or add any new features.

akheron:
Ok, LGTM.

PR #446, opened 2018-11-09
Create JSON_RAW type, implement helpers and test coverage.

coreyfarrell:
Related to issue #10
---
This is just an initial implementation of the json_raw type.  This patch does not include any changes to the parser, the idea is that one of the patches which allow numbers to not be parsed would be updated to store values in json_raw instead of using json_string with a flag.  The main goal is to avoid expanding the memory footprint of json strings which are very widely used.

apiref.rst needs updating to document the new functions.  I'll do that once it's decided that we want to proceed with this change.

### DIFF
--- a/src/value.c
+++ b/src/value.c
@@ -634,8 +634,9 @@ static json_t *json_array_deep_copy(const json_t *array)
 
 /*** string ***/
 
-static json_t *string_create(const char *value, size_t len, int own)
+static json_t *string_create(const char *value, size_t len, int own, json_type type)
 {
+    /* This function is used for JSON_STRING and JSON_RAW. */
     char *v;
     json_string_t *string;
 
@@ -655,7 +656,7 @@ static json_t *string_create(const char *value, size_t len, int own)
         jsonp_free(v);
         return NULL;
     }
-    json_init(&string->json, JSON_STRING);
+    json_init(&string->json, type);
     string->value = v;
     string->length = len;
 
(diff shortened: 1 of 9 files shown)


src/value.c
```diff
@@ -957,6 +959,110 @@ json_t *json_null(void)
...
+    return json_rawn(value, strlen(value));
+}
+
+json_t *json_rawn(const char *value, size_t len)
+{
+    if(!value || !utf8_check_string(value, len, 1))
+        return NULL;
+
+    return json_rawn_nocheck(value, len);
+}
```
akheron: Should we check that the raw value starts with a digit or one of `t`, `f`, `n`, `"`, `[`, `{`? Or should there even be a validation step that involves `json_loads` with `JSON_DECODE_ANY` flag on?

If not, should funtions for creating `JSON_RAW` values be private, at least in the beginning, so that it would be harder for users to create invalid JSON accidentally?

akheron: This of course assumes that there will be a decoding flag to e.g. decode all numbers as `JSON_RAW`.

coreyfarrell: You've given me the thought that `json_dump_raw` might be useful - something that works the same as json_dumps but returns a json_t containing a raw value.  Then json_raw constructors could be made private but end users could still create arbitrary raw objects and link them to other json_t objects/arrays.

I'm not sure about having these functions perform validation.  I guess for now I'm leaning towards making the create/set functions private?

coreyfarrell: I've dropped the json_raw constructors and setters, replaced with `json_dump_raw` and `json_dump_raw_new`.  The point of `json_dump_raw_new` is to support constructs such as:
```c
json_t *raw = json_dump_raw_new(json_pack("{}"), 0);
```

I think the 'create from string' functions should not be created until needed by `json_loads`, and at that point it will be up to json_loads to validate input even when flags tell it to keep certain input raw.

Issue #444, opened 2018-11-05
Core dumped on json_pack


Grindv1k:
I am doing some automated testing using this library and came across a core dump.

A minimal example:

```c
#include <jansson.h>

int main(int argc, char const *argv[])
{
    const char json_format[6] = "[1s";
    const char json_data[10]  = "Hi";

    json_pack(json_format, json_data);

    return(0);
}
```

My terminal output is 

```
*** Error in `./bug03': munmap_chunk(): invalid pointer: 0x000000000040077a ***
Aborted (core dumped)
```

and I believe this can be traced to  `jansson/src/memory.c:36` if I've understood my tooling output correctly.

As I am doing automated testing I do not actually know your API, so you'll have to tell me if this is a real bug or a false positive (i.e. something the user simply should not do).

coreyfarrell:
This is a bug in `pack_unpack.c:pack_string()`.  Although your format string is malformed this should not cause a segmentation fault, it should simply cause json_pack to return NULL.

json_pack format documentation: https://jansson.readthedocs.io/en/2.11/apiref.html#building-values

ploxiln:
I'm not sure if it's practical for `json_pack()` to ensure the proper number of arguments, and their types (which is OK if the format string is always strictly controlled by the programmer, typically as just a string literal argument).

coreyfarrell:
I said this specific case should not cause a segmentation fault.  The issue is that pack_string is incorrectly trying to free the string "Hi" which is impossible.

You are correct though we often cannot know of incorrect argument types, this is up to the programmer to perform testing/reviews on their code.

PR #448, opened 2018-11-25
jansson 2.12

coreyfarrell:
This is a bit of a grey area for 2.12.0 vs 2.11.1.  I decided to go for 2.12 since this includes the addition of function attributes which can be viewed as API changes (compile warnings added in certain situations).  The so-version only gets a revision bump as no ABI changes are included - a program compiled against 2.12 will run on a system where 2.11 is installed.  The primary reason I want to do the release is to provide a version with #444 as this fixes a regression in 2.11.  In theory I could cherry-pick that to create a v2.11.1 but I don't see any reason to hold back the other changes.

I've completed all steps from https://github.com/akheron/jansson/wiki/Releasing up to the commit.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,43 @@
+Version 2.12
+
+Released 2018-11-25
+
+* Bug fixes:
+
+  - Fix error message in `json_pack()` for NULL object (#409).
+
+  - Avoid invalid memory read in `json_pack()` (#421).
+
+  - Call va_end after va_copy in `json_vsprintf()` (#427).
+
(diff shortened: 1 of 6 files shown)


akheron:
v2.12 released, thanks a lot!

coreyfarrell:
Thanks!  Should I push the `v2.12.0` tag to a `2.12` branch like the other release versions have?  Maybe in the future I should have just pushed my 2.12 branch directly to your repo instead of my fork, I'm just used to working from my fork then using PR's to push things here.

SamyDgz (approved):

SamyDgz (approved):

PR #452, opened 2019-01-11
Add json_object_update_recursive()

hmpcabral:
Support merging values nested within objects. For instance, merging:

```json
    {
	"foo": 1,
	"bar": {
	    "baz": 2
	}
    }
```

with

```json
    {
	"bar": {
	    "baz": 3
	}
    }
```

results in

```json
    {
	"foo": 1,
	"bar": {
	    "baz": 3
	}
    }
```

instead of overwriting the value for the bar key.

### DIFF
--- a/doc/apiref.rst
+++ b/doc/apiref.rst
@@ -689,6 +689,12 @@ allowed in object keys.
 
    .. versionadded:: 2.3
 
+.. function:: int json_object_update_recursive(json_t *object, json_t *other)
+
+   Like :func:`json_object_update()`, but object values in *other* are
+   recursively merged with the corresponding values in *object* if they are also
+   objects, instead of overwriting them. Returns 0 on success or -1 on error.
+
 .. function:: json_object_foreach(object, key, value)
 
    Iterate over every key-value pair of ``object``, running the block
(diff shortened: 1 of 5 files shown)


coreyfarrell:
Thank you for this contribution.  I'd like to point you to #457 to mention that this new function also suffers from risk of infinite loop if it encounters a circular reference.  Potentially an issue for `object` and `other`.  Also concerning to me - what happens if `object` and `other` both reference the same sub-object, or if `other` contains two references to the same object?  As is this function risks editing part of `other` while updating `object`.

I think I'd want to see this split into public facing and private function.  The public function would check for both arguments being objects and check that no object members are referenced more than once between `object` and `other`.

harkue:
"json_deep_copy infinite loop" has been resolved in PR #490 .

coreyfarrell:
Thanks for the ping on this issue, if this is updated to incorporate the loop checks from #490 I can take another look.

AllenX2018:
I can create a pr based on this one incorporating loop check from #490 if needed.

AllenX2018:
@coreyfarrell Do you mean adding new commits to @hmpcabral's branch? I'm afraid I can't do that since I don't have write access to his fork.

ploxiln:
You can duplicate his branch into your own fork, and then add to it there.

ploxiln:
(and the commits will retain the current author metadata)

AllenX2018:
@ploxiln Thanks! I'll try

akheron:
Fixed in #505

Issue #453, opened 2019-01-14
Jansson Generating Documentation


AmeyaVS:
Hello,

Currently I am trying to build the Jansson documentation.
But even after installing the following library using `pip`: [http://www.sphinx-doc.org/en/master/](http://www.sphinx-doc.org/en/master/)

The CMake build is unable to locate the binary, and its version after fiddling with `FindSphinx.cmake`.

How should I setup the environment for generating the documentation?

Currently the `README.md` points to [http://sphinx.pocoo.org/](http://sphinx.pocoo.org/).
Which doesn't looks correct.

Does anyone has any idea on setting the documentation build? Till then I have to disable documentation generation.

-Ameya Vikram Singh

ploxiln:
You probably mean it can't find the binary `sphinx-build` - you need wherever pip puts it to be added to your `PATH`. For example, it might be at `/usr/local/bin/sphinx-build` or it might be in `$HOME/.local/bin/sphinx-build`. In the former case, if you do `export PATH=/usr/local/bin:$PATH` then cmake should be able to find `sphinx-build`.

If you paste here your `pip install` command and its output, the output of `pip --version`, and the `cmake` error output, I may be able to help more.

AmeyaVS:
Hello @ploxiln ,

The `sphinx-build` is available on the `PATH` environment variable.
Currently it is located at `~/.local/bin/sphinx-build`.

And here is the output for the command `sphinx-build --version`:
```sh
$ sphinx-build --version
sphinx-build 1.8.3
$ 
```

ploxiln:
so what's the cmake error

ploxiln:
hmm yeah it does look like the cmake version parsing of sphinx is out-of-date:

https://github.com/akheron/jansson/blob/v2.12/cmake/FindSphinx.cmake#L244-L264

it does `sphinx-build -h` and expects "Sphinx vX.Y.Z" but the help output no longer includes that. It should instead do `sphinx-build --version` and the output is "spinx-build X.Y.Z"

AmeyaVS:
Thanks @ploxiln for the hints.
I was close to fixing it but probably being too late into the night I could not see the regex expression.
I have created a pull-request with the fixes.

-Ameya Vikram Singh

AllenX2018:
> Thanks @ploxiln for the hints.
> I was close to fixing it but probably being too late into the night I could not see the regex expression.
> I have created a pull-request with the fixes.
> 
> -Ameya Vikram Singh

[...]

PR #454, opened 2019-01-15
Updated the Version detection for sphinx-build

AmeyaVS:
I have updated the `FindSphinx.cmake` to better align with the `sphinx-build` command line tool.

Faced the issue while building in an updated environment: #453

### DIFF
--- a/cmake/FindSphinx.cmake
+++ b/cmake/FindSphinx.cmake
@@ -246,13 +246,13 @@ if (Sphinx-build_EXECUTABLE)
   # will include the Sphinx version information
   if (Sphinx_PYTHON_EXECUTABLE)
     execute_process (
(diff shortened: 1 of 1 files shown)


ploxiln:
I did some quick experiments with a few older versions of Sphinx and observed that:

  * sphinx 1.7 and 1.8 have the version output format like "sphinx-build 1.7.2"
  * sphinx 1.2 through 1.6 have the version output format like "Sphinx (sphinx-build) 1.2.2"
  * sphinx 1.1 and before do not have a "--version" flag, but it causes the help output like "-h" does which includes version like "Sphinx v1.0.2"

It looks like Jansson's main CMakeLists.txt has two checks for sphinx version, where it just checks that sphinx version is at least 1.0, here's one:

https://github.com/akheron/jansson/blob/v2.12/CMakeLists.txt#L424

Given that version 1.0 is so old, I'd actually suggest just removing all the sphinx version parsing and checking. But we'll see what the maintainers think :)

akheron:
Fixed in #502

PR #459, opened 2019-01-29
Add JSON_FRACTIONAL_DIGITS encoding flag

pmahoney:
The JSON_FRACTIONAL_DIGITS encoding flag, when combined with the
JSON_REAL_PRECISION flag, causes the precision to indicate the number
of digits after the decimal (fractional digits) rather than the total
number of significant digits. Specifically, it changes the `printf`
conversion specifier from `g` to `f`.

---

A feature such as this would be useful for some use cases we have, such as reducing JSON size in a large output where high precision is not needed (but for various reasons, the %f conversion is preferable to %g for this particular case).

Is there any interest in incorporating this feature into jansson? This PR is an example of the feature I want; I'd be happy to modify it to better fit naming schemes or anything else.

Thank you.


m1r4ge:
+1

PR #461, opened 2019-02-15
Add location information to decoded JSON objects

n0-1:
Aid users in printing semantic errors in JSON input by allowing them to
point to syntactically valid JSON objects causing the semantic error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -509,6 +509,7 @@ if (NOT JANSSON_WITHOUT_TESTS)
          test_load
          test_loadb
          test_load_callback
+         test_location
          test_number
          test_object
          test_pack
--- a/doc/apiref.rst
+++ b/doc/apiref.rst
@@ -1257,6 +1257,12 @@ macros can be ORed together to obtain *flags*.
 
    .. versionadded:: 2.6
 
+``JSON_STORE_LOCATION``
+   Add location information to decoded :type:`json_t` objects. See
+   :ref:`apiref-location-information` for details.
+
+   .. versionadded:: 2.13
+
 Each function also takes an optional :type:`json_error_t` parameter
 that is filled with error information if decoding fails. It's also
 updated on success; the number of bytes of input read is written to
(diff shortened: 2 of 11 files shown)


coreyfarrell:
I'm sorry I do not feel that this is appropriate for jansson.  This adds 16 bytes to every single json_t for something that is only filled in by the decode process.  This means that software which heavily uses `json_pack` will start using more memory, especially when dealing with many complex objects.  Additionally this potentially represents a breaking change - json_t is a public structure and I'm unsure the ABI ramifications of changing it's size.

In theory the JSON parser could be refactored to support a callback based interface, similar to SAX for example.  This could be accomplished without a breaking change to give access to lexer internals so you'd be able to create associations between json_t objects and the lexer state.  This could also potentially help us move towards a resume-able parser, though I'm just speculating as it would depend on the implementation.

n0-1:
Hi Corey,

On Sat, Feb 16, 2019 at 12:23:57PM +0000, Corey Farrell wrote:
> I'm sorry I do not feel that this is appropriate for jansson.  This adds 16 bytes to every single json_t for something that is only filled in by the decode process.  This means that software which heavily uses `json_pack` will start using more memory, especially when dealing with many complex objects.  Additionally this potentially represents a breaking change - json_t is a public structure and I'm unsure the ABI ramifications of changing it's size.

I see. So in quintessence:

* Don't bloat data structures pointlessly
* Don't change json_t to avoid breaking ABI.

> In theory the JSON parser could be refactored to support a callback based interface, similar to SAX for example.  This could be accomplished without a breaking change to give access to lexer internals so you'd be able to create associations between json_t objects and the lexer state.  This could also potentially help us move towards a resume-able parser, though I'm just speculating as it would depend on the implementation.

In my opinion this isn't what jansson should evolve into. In a separate
API maybe, but event-based parsers tend to be bloaty and complex. Both
aspects that gladly can't be attributed to jansson.

I'd rather try a second attempt lacking the flaws you mentioned:

One measure could be to have a struct location and merely put pointers
to it into json_* structures. This way their size increases only by
eight bytes (on 64bit systems) and since allocation happens only when
decoding, there's no computational overhead in other cases.

Instead of modifying json_t directly, I could enhance json_*_t
structures instead. This should allow for ABI compatibility.
Alternatively, I could introduce something like json_loaded_t which
wraps the types and location information.

Would something like that work for you?

Thanks, Phil

n0-1:
This is a respin of my original PR, changes are:

- Put location info into private JSON objects to keep ABI stable.
- Store location only if JSON_STORE_LOCATION flag was given during decoding.
- Introduce json_location_t holding location info, dynamically allocated to reduce memory footprint impact if JSON_STORE_LOCATION is not used.
- Added code to convert singleton objects into refcounted ones so they can hold location information if requested.
- Introduce json_get_location() to safely retrieve location info if available.
- Adjust testcase to give more detailed error output using fail_args().

n0-1:
Rebased my feature branch onto current master branch, resolving a minor conflict in src/jansson.h due to context change.

n0-1:
Fixed for failing CI checks:
- Add new json_get_location API function to src/jansson.def
- Rename private functions to be prefixed with jsonp.
- Dropped stdbool.h include from test/suites/api/test_location.c and changed code to use int 0/1 instead to prevent compile failures in MSVC.

n0-1:
One more try, this time covering for crufty MSVC, merely fixing test code:
- Don't use snprintf().
- Avoid non-constant expressions when declaring arrays.

akheron:
Hi, and thanks for your continued effort on this PR. Unfortunately we don't see this as an important feature that should be included in Jansson. The library just turned 10 years old, and we don't recall anyone requesting this feature before, so it seems that the vast majority of the users is fine without it.

PR #465, opened 2019-03-08
Add runtime version checking functions

seanbright:
This patch adds two new exported functions:
    
* `jansson_version_str` - Returns a human-readable version number
* `jansson_version_cmp` - Returns an integer less than, equal to, or greater
  than zero if the runtime version of Jansson is found, respectively, to be
  less than, to match, or be greater than the provided major, minor, and micro.

### DIFF
--- a/doc/apiref.rst
+++ b/doc/apiref.rst
@@ -58,6 +58,25 @@ the library:
       /* Code specific to version 1.3 and above */
       #endif
 
+Additionally, there are functions to determine the version of Jansson at
+runtime:
+
+.. function:: const char *jansson_version_str()
+
+   Return the version of the Jansson library, in the same format as
+   the ``JANSSON_VERSION`` preprocessor constant.
+
+   .. versionadded:: 2.13
+
+.. function:: int jansson_version_cmp(int major, int minor, int micro)
+
+   Returns an integer less than, equal to, or greater than zero if
+   the runtime version of Jansson is found, respectively, to be less
+   than, to match, or be greater than the provided *major*, *minor*, and
+   *micro*.
+
+   .. versionadded:: 2.13
+
 ``JANSSON_THREAD_SAFE_REFCOUNT``
   If this value is defined all read-only operations and reference counting in
   Jansson are thread safe.  This value is not defined for versions older than
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -20,7 +20,8 @@ libjansson_la_SOURCES = \
 	strconv.c \
 	utf.c \
 	utf.h \
-	value.c
+	value.c \
+	version.c
 libjansson_la_LDFLAGS = \
 	-no-undefined \
 	-export-symbols-regex '^json_' \
--- a/src/jansson.def
+++ b/src/jansson.def
@@ -72,4 +72,6 @@ EXPORTS
     json_vunpack_ex
     json_set_alloc_funcs
     json_get_alloc_funcs
+    jansson_version_str
+    jansson_version_cmp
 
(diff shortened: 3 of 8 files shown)


seanbright:
A couple of notes that I didn't think belonged in the commit message:

1. I used the `jansson_` prefix instead of `json_` to prevent potential naming conflicts with other libraries.
2. I put my copyright in the new `version.c`

I have no problems changing the prefix or disclaiming copyright if that is required for inclusion.

akheron:
What's the use case for the new functions?

coreyfarrell:
The Asterisk 16+ configure script checks for jansson >= 2.11, rejects the build if it's not used.  We've had a user report [ASTERISK-28329](https://issues.asterisk.org/jira/browse/ASTERISK-28329) which is showing an error:
```
Error building JSON from '{s: o?, s: o, s: O?}': Expected format 's', got '?'.
```
This indicates to us that although they built against jansson 2.11+ they're actually linking with 2.7 or older at runtime.  The current plan is to perform a [startup check](https://gerrit.asterisk.org/#/c/asterisk/+/11112/3/main/json.c) to verify that the newer format strings work but in the future it might be better for us to simply check that the runtime version meets the minimum requirement.

My only thought on naming is that if `json_*` conflicts with another library we already conflict, so I lean toward just sticking with the existing naming convention.  Maybe `json_get_version` and `json_get_version_str`?  I don't have a very strong opinion on this.  

Maybe we would be better served by `json_cmp_version(int major, int minor, int micro)` instead of the function that returns the version integer?  This way Asterisk wouldn't need to concern itself with encoding of the JANSSON_VERSION_HEX value.  The function that returns the version string would still be nice so we could log, so when we require jansson-2.15:
```c
if (json_cmp_version(2, 15, 0) < 0) {
  ast_log(LOG_ERROR, "Expected jansson-2.15 but found %s\b", json_get_version_str());
  exit(1);
}
```

@seanbright `doc/apiref.rst` needs to mention the new functions.

I have no opinion about the copyright thing, that's a question for @akheron.

akheron:
I see. I thought that dynamic linker would perform such version checks at program startup, as jansson 2.11 has "added interfaces" in libtool terms compared to jansson 2.7.

coreyfarrell:
@akheron it does not.  I just verified in a Fedora 28 docker container. I compiled Asterisk 16 against the system jansson-2.11, then downgraded to jansson-2.7 and ran Asterisk.  No dynamic linker errors.  I added our patch which uses a 'new' json_pack format and it detected the lack of support for `"o*".

ploxiln:
Typically you bump the so-name when making a backwards-incompatible change, but not when adding a new feature in a backwards-compatible way. So if you update the library with new backwards-compatible features, the same old application binary can still link to it.

But the other way around - compiling an application against a newer version of the library and then running it with an older version of the library - is much less assured. Often, new features come as new symbols (e.g. functions) and the dynamic linker fails to find them. In some cases it's some other kind of option value and only fails later in runtime, I think this is to be expected.

I think a dynamic-linked get-version function is a fine idea.

akheron:
Ok, let's add what's needed to solve this.

I don't feel strongly about whether it should be a comparison function or one that returns the hex encoded version.

I actually prefer the `jansson_` prefix for these functions, because `json_` could be understood as the version of JSON, or something. I don't feel strongly about this either.

The copyright notice is ok. It would feel weird to have my copyright on a file I didn't write :)

Apiref docs need to be updated in any case.

seanbright:
I _think_ I've addressed all comments. Let me know if there is anything else.

doc/apiref.rst
```diff
@@ -58,6 +58,21 @@ the library:
       /* Code specific to version 1.3 and above */
       #endif
 
+Additionally, as of ``2.13``, there are functions to determine the version of
+Jansson at runtime:
```
akheron: There's `.. versionadded:` directive that you should use for documenting the version in which a feature was added. See e.g. the docs of `json_boolean_value()`

akheron: Otherwise the PR looks good!

PR #468, opened 2019-03-20
Make CMAKE_DEBUG_POSTFIX truly optional

### LINKED ISSUE #385: CMake remove CMAKE_DEBUG_POSTFIX or add Option
like that
if(NOT JSON_NODEBUG_POSTFIX)
    if (WIN32)
       set(CMAKE_DEBUG_POSTFIX "_d")
    endif (WIN32)
endif

KOConchobhair:
The missing piece to fix  #385

Add DEFINED to the conditional to allow the user to set CMAKE_DEBUG_POSTFIX to blank which will disable the _d suffix on Windows builds. Othwerise, the default jansson cmake behavior is unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ set(JANSSON_TEMP_DIR ${CMAKE_CURRENT_BINARY_DIR}/tmp)
 # Give the debug version a different postfix for windows,
 # so both the debug and release version can be built in the
 # same build-tree on Windows (MSVC).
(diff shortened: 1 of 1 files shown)


harkue:
LGTM!  It is a better way to set the value of CMAKE_DEBUG_POSTFIX.
It's also seen in [curl](https://github.com/curl/curl) project like [73a3b3e0](https://github.com/curl/curl/pull/2599/commits/73a3b3e05b9f90e3de54efc85e64a234dacd7a3f)

JosiahWI (approved):
Thanks for fixing this!

PR #470, opened 2019-04-05
Fix warning in src/load.c

CoRfr:
With gcc 8.2 there is the following warning:

```
jansson/src/load.c: In function 'error_set':
jansson/src/load.c:118:29: warning: ' near '' directive output may be truncated writing 7 bytes into a region of size between 1 and 160 [-Wformat-truncation=]
                          "%s near '%s'", msg_text, saved_text);
                             ^~~~~~~
jansson/src/load.c:117:17: note: 'snprintf' output 9 or more bytes (assuming 168) into a destination of size 160
                 snprintf(msg_with_context, JSON_ERROR_TEXT_LENGTH,
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                          "%s near '%s'", msg_text, saved_text);
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
jansson/src/load.c:135:29: warning: ' near end of file' directive output may be truncated writing 17 bytes into a region of size between 1 and 160 [-Wformat-truncation=]
                          "%s near end of file", msg_text);
                             ^~~~~~~~~~~~~~~~~
jansson/src/load.c:134:17: note: 'snprintf' output between 18 and 177 bytes into a destination of size 160
                 snprintf(msg_with_context, JSON_ERROR_TEXT_LENGTH,
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                          "%s near end of file", msg_text);
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Adjusting the size of buffer to give enough room to snprintf.

### DIFF
--- a/src/load.c
+++ b/src/load.c
@@ -89,7 +89,7 @@ static void error_set(json_error_t *error, const lex_t *lex,
 {
     va_list ap;
     char msg_text[JSON_ERROR_TEXT_LENGTH];
-    char msg_with_context[JSON_ERROR_TEXT_LENGTH];
+    char msg_with_context[JSON_ERROR_TEXT_LENGTH + 20];
 
     int line = -1, col = -1;
     size_t pos = 0;
@@ -114,9 +114,8 @@ static void error_set(json_error_t *error, const lex_t *lex,
         if(saved_text && saved_text[0])
         {
             if(lex->saved_text.length <= 20) {
-                snprintf(msg_with_context, JSON_ERROR_TEXT_LENGTH,
+                snprintf(msg_with_context, sizeof(msg_with_context),
                          "%s near '%s'", msg_text, saved_text);
-                msg_with_context[JSON_ERROR_TEXT_LENGTH - 1] = '\0';
                 result = msg_with_context;
             }
         }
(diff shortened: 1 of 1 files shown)


coreyfarrell:
configure.ac disables the format-truncation warning.  I'm guessing you are using cmake which does not currently disable that warning.  So this needs to be fixed in the cmake build system, I'm not sure how to accomplish that.  Support for the `-Wno-format-truncation` gcc flag needs to be detected as is done in configure.ac, I'm not sure how to accomplish that for cmake.

This patch suppresses the warning in these places by shifting the potential truncation to happen inside `jsonp_error_set` which does not produce a warning because it's receiving the over-length text by pointer.  So this patch expands the stack memory usage without accomplishing anything (the same exact truncation will still occur).  I realize 20 bytes is a small increase but jansson targets embedded platforms so my default position is against wasting stack space.

coreyfarrell:
@CoRfr can you confirm that you are not using `./configure` to build your library?

CoRfr:
@coreyfarrell , I confirm that we are not using the `./configure` script to build the library, which is why these warnings show up.

coreyfarrell:
@CoRfr thanks for letting me know.  The CMake build scripts needs to be patched to detects if the compiler supports/tolerates the `-Wno-format-truncation` argument, add it if appropriate to the CFLAGS used to compile.  [configure.ac](https://github.com/akheron/jansson/blob/master/configure.ac#L153) does this already, I'm not familiar enough with CMake to patch that.

AllenX2018:
This can be done by ```CheckCCompilerFlag``` module in cmake build system. Use ```check_c_compiler_flag``` command to detect if the c compiler support ```-Wno-format-truncation``` argument. If the compiler support it, append it to ```CMAKE_C_FLAGS``` variable. Then the warning will be suppressed.

coreyfarrell:
Closing in favor of #489

PR #477, opened 2019-06-12
jansson.h: suppress unused-function warnings

brong:
Without this pragma, building with clang and -Wall I get the following:

```
In file included from imap/mboxevent.c:53:
/usr/local/cyruslibs/include/jansson.h:140:6: error: unused function 'json_decrefp' [-Werror,-Wunused-function]
void json_decrefp(json_t **json)
     ^
/usr/local/cyruslibs/include/jansson.h:186:41: error: unused function 'json_error_code' [-Werror,-Wunused-function]
static JSON_INLINE enum json_error_code json_error_code(const json_error_t *e) {
                                        ^
/usr/local/cyruslibs/include/jansson.h:228:5: error: unused function 'json_object_set' [-Werror,-Wunused-function]
int json_object_set(json_t *object, const char *key, json_t *value)
    ^
/usr/local/cyruslibs/include/jansson.h:234:5: error: unused function 'json_object_set_nocheck' [-Werror,-Wunused-function]
int json_object_set_nocheck(json_t *object, const char *key, json_t *value)
    ^
/usr/local/cyruslibs/include/jansson.h:240:5: error: unused function 'json_object_iter_set' [-Werror,-Wunused-function]
int json_object_iter_set(json_t *object, void *iter, json_t *value)
    ^
/usr/local/cyruslibs/include/jansson.h:255:5: error: unused function 'json_array_set' [-Werror,-Wunused-function]
int json_array_set(json_t *array, size_t ind, json_t *value)
    ^
/usr/local/cyruslibs/include/jansson.h:261:5: error: unused function 'json_array_append' [-Werror,-Wunused-function]
int json_array_append(json_t *array, json_t *value)
    ^
/usr/local/cyruslibs/include/jansson.h:267:5: error: unused function 'json_array_insert' [-Werror,-Wunused-function]
int json_array_insert(json_t *array, size_t ind, json_t *value)
    ^
```

And ditto everywhere else jansson.h is included.  This patch has no effect on the code built, but allows callers to turn on full warnings safely with clang.


ploxiln:
For most versions of gcc and clang that I have seen, the `inline` directive prevents that warning, usually that warning is only omitted for `static` _without_ `inline`.

coreyfarrell:
Agreed with @ploxiln this should not be an issue if `inline` is enabled.  How are you configuring your libjansson?  Please post your jansson_config.h to a gist, in particular I'm interested in seeing what you have for `#define JANSSON_INLINE`.

brong:
This is our entire build script, with a fresh git checkout:

```
  git clean -f -x -d
  autoreconf -v -i
  ./configure --enable-silent-rules --prefix=$PREFIX
  make -j 8
  sudo make install
```

When I just tried that on my laptop now, I got a `jansson_private_config.h` but no `jansson_config.h` and my `jansson_private_config.h` did not match JANSSON_INLINE at all.

I will attempt to recreate the scenario and grab you the files from the build.

brong:
> For most versions of gcc and clang that I have seen, the `inline` directive prevents that warning, usually that warning is only omitted for `static` _without_ `inline`.

Sorry I missed this - perils of filing pull requests just before traveling!

coreyfarrell:
src/jansson_config.h is created when `./configure` succeeds.  If this file were not created then no part of jansson would compile.

brong:
Ahh right - it's the "src/" bit that I missed.  Ok, it's at https://gist.github.com/brong/af81aedc86b709ffaf1684c13d663ae0

I am trying to recreate by removing clang to build cyruslibs, then adding clang again before building cyrus.

brong:
Anndd.. there it is :)

```In file included from imap/mboxevent.c:53:
/usr/local/cyruslibs-jansson-test/include/jansson.h:140:6: error: unused function 'json_decrefp' [-Werror,-Wunused-function]
void json_decrefp(json_t **json)
```

This happens when cyruslibs was built with gcc, and cyrus is being built with clang.

coreyfarrell:
Your compiler/preprocessor is using `static inline` so this should result in the unused function being ignored.  Are you using older gcc or clang?

If needed I'm OK with adding `JANSSON_ATTRS(unused)` directly to the inline functions (this should explicitly suppress the warning for each function).  I'm not in favor of using the pragma method.

PR #478, opened 2019-06-27
add CMake option JANSSON_OUTPUT_DIRECTORY

irov:
Issue #478

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,11 +68,15 @@ if (UNIX)
    option(JANSSON_COVERALLS_UPLOAD "Upload coverage info to Coveralls (Only works via Travis)" ON)
 endif ()
 
-# Set some nicer output dirs.
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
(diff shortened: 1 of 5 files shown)


coreyfarrell:
@Mephistophiles @isaachier I'm really not qualified to determine if most changes to the cmake build is appropriate, tagging you since you've both made changes to cmake.  Please let me know if this PR is a good idea.  In particular I don't want for us to create custom variables which duplicate built-in functionality.

Mephistophiles:
Maybe @irov wanted CMAKE_INSTALL_PREFIX ? 

Best cross-compile practices show buildroot package:
https://github.com/buildroot/buildroot/blob/cf9e8ebe47fc29895e7017660e88158a0197bb14/package/pkg-cmake.mk#L113

irov:
> Maybe @irov wanted CMAKE_INSTALL_PREFIX ?

Yes, it also needs to be done. _[I think this will also solve the problem, I would check the patch for myself]_

But I also wanted to be able to setup 
```
CMAKE_RUNTIME_OUTPUT_DIRECTORY 
CMAKE_LIBRARY_OUTPUT_DIRECTORY 
CMAKE_ARCHIVE_OUTPUT_DIRECTORY
```
My engine uses many dependencies, and they all allow me to manage these variables, or they do not set them

harkue:
IMO, this is a special case, and also we can solve the problem of using "./configure --prefix=DIR" to specify the install path.

PR #480, opened 2019-06-28
oss-fuzz: Create a fuzzing target for jansson

cmeister2:
Hi! I've been adding oss-fuzz support to some open-source projects (lz4, curl) and thought that I'd offer a fix for issue #479 raised by @Google-Autofuzz.

This takes the sample code they offered and puts it into your build tree. To actually get it into oss-fuzz you'll need to submit a patch in the oss-fuzz project; if this is something you're interested in I have a code branch ready to submit (just needs a contact email address for any bugs that it finds).

re: the raw Make commands: Annoyingly they've recently changed the behaviour of the LIB_FUZZING_ENGINE variable (it used to be a file, now its value is "-fsanitizer=fuzzer" (so now is a flag). The Make commands here mean that you can build a testcase loader without having to have that set up.


coreyfarrell:
Hi @cmeister2 I've just returned from being AFK for nearly two weeks.  I'm a bit buried at the moment but promise I will take a look at this as soon as I can.

cmeister2:
@coreyfarrell thanks for getting back to me! No urgent rush here obviously :smile:

cmeister2:
@coreyfarrell any movement here? I've got some spare time today to make any markups.

coreyfarrell:
@cmeister2 sorry I've been so busy.  One thing about the current PR, I'd prefer if ossfuzz were moved to test/ossfuzz.  Also any reason we can't put a Makefile.am inside the ossfuzz directory instead of modifying the root Makefile?

Beyond that I'm interested in seeing the results of this.

Google-Autofuzz:
As a side note, here is a slightly improved fuzzer:

```C
#include <stdint.h>
#include <sys/types.h>

#include "jansson/src/jansson.h"

extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
  json_error_t error;

  if (size < sizeof(size_t) + sizeof(size_t))
    return 0;

  size_t load_flags = *(const size_t*)data;
  data += sizeof(size_t);
  size_t dump_flags = *(const size_t*)data;
  data += sizeof(size_t);

  size -= sizeof(size_t) + sizeof(size_t);

  const char* text = reinterpret_cast<const char *>(data);
  json_t* jobj = json_loadb(text, size, load_flags, &error);

  if (!jobj)
    return 0;

  char* out = json_dumps(jobj, dump_flags);

  if (out)
    free(out);
  if (jobj)
    json_decref(jobj);

  return 0;
}
```

cmeister2:
Have just pushed another commit which works locally. It checks the value of LIB_FUZZING_ENGINE in configure.ac and sets conditionals accordingly, which are then used in Makefile.am to have the correct behaviour. I also added @Google-Autofuzz's updated fuzzing code.

coreyfarrell:
@cmeister2 A few issues:

1. Looks like test/ossfuzz/json_load_fuzzer is a compiled progam which should not be committed.
2. Can the changes to the root `Makefile.am` be removed now?
3. Is the program supposed to be named `json_load_dump_fuzzer` or `json_load_fuzzer`?  Whichever filename you want to output please make sure that it is not committed and is listed in `.gitignore`.
4. It seems like the `test/ossfuzz` is always built even for normal builds?

I suspect this worked locally for you because test/ossfuzz/json_load_fuzzer was committed / left behind.  Please retest after deleting that file / from a clean checkout where `git status --ignored` doesn't list any files.

cmeister2:
Oops, how embarrassing. I'll fix it tomorrow morning!

My automake-fu is not yet strong enough to know how to do optional
programs; I'll look into it if it's going to be a problem?

cmeister2:
@coreyfarrell:

1. Looks like test/ossfuzz/json_load_fuzzer is a compiled progam which should not be committed.

Now removed!

2. Can the changes to the root Makefile.am be removed now?

Yes, now removed.

3. Is the program supposed to be named json_load_dump_fuzzer or json_load_fuzzer? Whichever filename you want to output please make sure that it is not committed and is listed in .gitignore.

json_load_dump_fuzzer is the new name, this is the code from the post above (which dumps as well as loads!)

4. It seems like the test/ossfuzz is always built even for normal builds?

Yes, that was the case. I've now hidden this behind "--enable-ossfuzzers", obviously can name this whatever.

cmeister2:
@coreyfarrell Would also be good at some point to get a contact gmail account for the OSS Fuzz reports - can you point me at a good email address to use?

coreyfarrell:
For now please point the reports to `git@cfware.com`.  Will the emails contain sensitive information / is that why you asked for a gmail address?  In any case cfware.com email is hosted by gmail so this should enable ossfuzz to send me emails that do not travel unencrypted over the internet.

cmeister2:
@coreyfarrell Basically you need a google account to log into ossfuzz.com; that has to be added to the ossfuzz list.

I'll tag you on the ossfuzz PR so you can see what I mean!

SamyDgz (approved):

PR #484, opened 2019-07-22
One byte fix to fix OSS Fuzz AFL support

cmeister2:
"x$LIB_FUZZING_ENGINE" is not the correct filename to test for - remove the x!


cmeister2:
(@coreyfarrell this will fix the build failure in https://github.com/google/oss-fuzz/pull/2622)

SamyDgz (approved):

PR #485, opened 2019-07-24
ossfuzz: Improvements to json_load_dump_fuzzer

cmeister2:
Add in diagnostics to json_load_dump_fuzzer so that if FUZZ_VERBOSE is
defined in the environment, additional debug information about the
current test case is output to stderr.

Replace json_dumps by json_dump_callback so that memory is not exhausted
by excessively nested json objects.

---

This should work around https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=16031 (not yet publicly visible).


cmeister2:
@coreyfarrell as discussed, added in a dump_mode byte.

PR #487, opened 2019-07-24
ossfuzz: Zip up all "input" files to use as a seed corpus

cmeister2:
@Dor1s suggested that if we zip up all the "input" files from the test/suites directory, then that would be good to use as a seed corpus. I don't necessarily know if these will work as inputs (the directory structure is maintained here, and the inputs don't respect the fact that there's a certain number of header bytes used for command flags for fuzzer inputs) but in the hope that they are correct, here is a PR which zips up all the input files :smile:

### DIFF
--- a/test/ossfuzz/ossfuzz.sh
+++ b/test/ossfuzz/ossfuzz.sh
@@ -25,3 +25,6 @@ make
 
 # Copy the fuzzer to the output directory.
 cp -v test/ossfuzz/json_load_dump_fuzzer $OUT/
+
+# Zip up all input files to use as a test corpus
+find test/suites -name "input" -print | zip $OUT/json_load_dump_fuzzer_seed_corpus.zip -@


coreyfarrell:
I'm confused about this, would the google repository handle adding fuzzed load_flags, dump_mode and dump_flags?

How would this deal with the expected failure cases - see places where an `error` file exists next to `input` rather than an `output` file.  Maybe you would need to find all `output` files instead of input?

cmeister2:
I think it's just the inputs that are the interesting part, so I don't think we need the error files.

WRT the command flags, it's not clear. Knowing libfuzzer, I think it'll take the corpus and use fragments from the files to populate its database; whether this is good enough, I don't know.

coreyfarrell:
Ahh I just looked at json_load_dump_fuzzer.cc and see that it tolerates invalid input, just exits early.  I'll merge this and we'll see what it does.

Issue #441, opened 2018-10-09
__strtod


jwhere:
gcc --version
gcc.exe (MinGW.org GCC-6.3.0-1) 6.3.0
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

gcc -DHAVE_CONFIG_H=1 -s -static-libgcc -shared -o jansson.dll *.c
strconv.c: In function 'jsonp_strtod':
strconv.c:18:16: warning: implicit declaration of function '__strtod' [-Wimplicit-function-declaration]
 #define strtod __strtod
                ^
strconv.c:77:13: note: in expansion of macro 'strtod'
     value = strtod(strbuffer->value, &end);
             ^~~~~~

------------------------------------------------------------------------------------------------------

#include <stdio.h>
#include <jansson.h>

int main(int argc, char *argv[]) {
    json_t *json = json_loads("{\"c\":1,\"score\":2991.0}", 0, NULL);
    printf("%s\n", json_dumps(json, 0));
    return 0;
}

[output]
{"c": 1, "score": 1.0}

------------------------------------------------------------------------------------------------------

solution

strconv.c:

#include <assert.h>
#include <errno.h>
#include <stdio.h>
#include <string.h>
#include <math.h>
#ifdef __MINGW32__
#undef __NO_ISOCEXT /* ensure stdlib.h will declare prototypes for mingw own 'strtod' replacement, called '__strtod' */
#endif
#include "jansson_private.h"
#include "strbuffer.h"

/* need jansson_private_config.h to get the correct snprintf */
#ifdef HAVE_CONFIG_H
#include <jansson_private_config.h>
#endif

▽▽▽▽ REMOVE LINE 18 ▽▽▽▽
#ifdef __MINGW32__
#define strtod __strtod
#endif

AllenX2018:
It reproduced. If we use __strtod, it will result in 4 of 5 test suits(about precesion) failed.

By the way, find it a note in stdlib.h written as:
```c
_CRTIMP __cdecl __MINGW_NOTHROW  double strtod (const char *, char **);
.........

#if __USE_MINGW_ANSI_STDIO
/* Microsoft's strtod() and atof() implementations, (in MSVCRT.DLL),
 * mishandle infinities and NaNs; on the basis that this conditional
[...]

PR #498, opened 2019-10-08
fix issue #441: remove strtod definition in strconv.c

AllenX2018:
It already has "strtod" function in stdlib.h and because __USE_MINGW_ANSI_STDIO is desperated

### DIFF
--- a/src/strconv.c
+++ b/src/strconv.c
@@ -14,10 +14,6 @@
 #include <jansson_private_config.h>
 #endif
 
-#ifdef __MINGW32__
-#define strtod __strtod
-#endif
-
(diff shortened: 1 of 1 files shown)


akheron:
Could the `#undef` line near the top of the file also be removed now?

```
#undef __NO_ISOCEXT /* ensure stdlib.h will declare prototypes for mingw own 'strtod' replacement, called '__strtod' */
```

akheron:
Merged, thanks! Btw, you should add a comment when pushing a change, since github won't send any notification emails on pushes.

SamyDgz (approved):

PR #499, opened 2019-10-08
Add json_object_update_new, json_object_update_existing_new and json_object_update_missing_new

### LINKED ISSUE #281: [Feature Request] Can you provide a function similar to json_object_set_new for json_object_update
Cannot you provide a function json_object_update_new(json_t *object, json_t *other), This is useful when "other" is newly created and not used after the call. Thanks.

AllenX2018:
Fixes #281

### DIFF
--- a/src/jansson.h
+++ b/src/jansson.h
@@ -242,6 +242,30 @@ int json_object_iter_set(json_t *object, void *iter, json_t *value)
     return json_object_iter_set_new(object, iter, json_incref(value));
 }
 
+static JSON_INLINE
+int json_object_update_new(json_t *object, json_t *other)
+{
+    int ret = json_object_update(object, other);
+    json_decref(other);
+    return ret;
+}
+
+static JSON_INLINE
+int json_object_update_existing_new(json_t *object, json_t *other)
+{
+    int ret = json_object_update_existing(object, other);
+    json_decref(other);
+    return ret;
+}
+
+static JSON_INLINE
+int json_object_update_missing_new(json_t *object, json_t *other)
+{
+    int ret = json_object_update_missing(object, other);
+    json_decref(other);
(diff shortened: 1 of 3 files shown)


src/jansson.h
```diff
@@ -242,6 +242,30 @@ int json_object_iter_set(json_t *object, void *iter, json_t *value)
     return json_object_iter_set_new(object, iter, json_incref(value));
 }
 
+static JSON_INLINE
+int json_object_update_new(json_t *object, json_t *other)
+{
+    int ret = json_object_update(object, other);
+    json_decref(other);
```
akheron: I'm not entirely sure, but I think that you should call `json_decref` here only if `ret == 0`. Otherwise the `other` object will be freed if the `json_object_update` call fails. Could you check the semantics of other already existing `_new` functions to align the reference counting semantics with them?

akheron: The same applies to the other two added functions also.

akheron:
I edited the PR title and first comment so that
- the referenced issue is automatically closed when this PR is merged
- looking at merged PRs makes it easier to update the CHANGELOG by looking at the PR titles

src/jansson.h
AllenX2018: I think no matter the ```json_object_update``` call succeed or not, the ```other``` always need to be freed, since the reference of ```other``` will be stolen by this ```json_object_update_new``` function.
In fact, this function will be called like:
```C
json_object_update_new(object, json_pack("{s{ss}}", "foo", "bar", "baz"));
```

Here ```other```, created by calling ```json_pack```, is anonymous, which means it cannot be freed outside the ```json_object_update_new```

akheron: Ah, you're right.

PR #508, opened 2019-10-17
Format code with clang-format

akheron:
Also ensure that everything keeps correctly formatted by running clang-format in the CI and failing the build if any file is not formatted correctly.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,5 @@
 BasedOnStyle: LLVM
 AlignConsecutiveMacros: true
-AllowShortCaseLabelsOnASingleLine: true
+ColumnLimit: 90
 IndentCaseLabels: true
 IndentWidth: 4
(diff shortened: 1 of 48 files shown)


coreyfarrell:
One more thought, since CI will fail upon formatting issues maybe that should just be part of `make check` instead?  It's not awesome when you take the time to run `make && make check`, everything passes then CI rejects.

akheron:
Good observations! I added the `AllowShortCaseLabelsOnASingleLine` directive because I thought it generated a slightly smaller diff, but I actually didn't like it that much either :) I set the maximum line length to 90 now. `make check` now also runs the clang-format check, and the test is skipped if the user doesn't have clang-format installed. We should probably have a development section in readme to document this, and any other development practices if there's something worth mentioning.

Merging this will conflict with all the other open PRs. Do you think there's any PR that would be better to get ready before this onet?

coreyfarrell:
> Merging this will conflict with all the other open PRs. Do you think there's any PR that would be better to get ready before this onet?

Maybe #441?

PR #513, opened 2019-11-18
add oss-fuzz status badge in README

AllenX2018:
Add the oss-fuzz badge in README to see bugs/errors found by the OSS-Fuzz integration at a glance


coreyfarrell:
I'm not sure how I feel about this, specifically I don't like the target page.  The left panel lists a bunch of other projects with red exclamation marks.  I don't want to cause confusion for people who might not realize these errors are for other projects.

PR #517, opened 2019-12-15
Enable and fix additional warnings

danielkempenich:
Enable pedantic, strict prototypes, and treat warnings as errors

Resolve several cases of: function declaration isn’t a prototype
[-Werror=strict-prototypes]

Resolve: format ‘%p’ expects argument of type ‘void *’, but argument 4
has type ‘const json_t * {aka const struct json_t *}’ [-Werror=format=]

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -158,7 +158,14 @@ if test x$GCC = xyes; then
       [AC_MSG_RESULT(no)
       wnoformat_truncation=""])
 
-    AM_CFLAGS="-Wall -Wextra -Wdeclaration-after-statement -Wshadow ${wnoformat_truncation}"
+    AC_MSG_CHECKING(for -Wstrict-prototypes)
+    wstrict_prototypes="-Wstrict-prototypes"
+    AS_IF([${CC} -Wstrict-prototypes -Werror -S -o /dev/null -xc /dev/null > /dev/null 2>&1],
+      [AC_MSG_RESULT(yes)],
+      [AC_MSG_RESULT(no)
+      wstrict_prototypes=""])
+
+    AM_CFLAGS="-pedantic -Wall -Wextra -Wdeclaration-after-statement -Wshadow ${wnoformat_truncation} ${wstrict_prototypes}"
 fi
 AC_SUBST([AM_CFLAGS])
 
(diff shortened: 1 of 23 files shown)


configure.ac
```diff
@@ -158,7 +158,14 @@ if test x$GCC = xyes; then
...
 
-    AM_CFLAGS="-Wall -Wextra -Wdeclaration-after-statement -Wshadow ${wnoformat_truncation}"
+    AC_MSG_CHECKING(for -Wstrict-prototypes)
+    wstrict_prototypes="-Wstrict-prototypes"
+    AS_IF([${CC} -Wstrict-prototypes -Werror -S -o /dev/null -xc /dev/null > /dev/null 2>&1],
+      [AC_MSG_RESULT(yes)],
+      [AC_MSG_RESULT(no)
+      wstrict_prototypes=""])
+
+    AM_CFLAGS="-pedantic -Wall -Wextra -Werror -Wdeclaration-after-statement -Wshadow ${wnoformat_truncation} ${wstrict_prototypes}"
```
ploxiln: `-Wstrict-prototypes`, OK, sure. `-Werror`, not a good idea.

By all means, fix compiler warnings. I do! I keep my compiles clean.

But don't turn warnings into errors. This causes complete failure when just switching to a newer compiler, or a different compiler, or when comparing an older version of the library to a newer version of the library, while compiling both versions with a new compiler. If a linux distro or a self-contained project wants to update the compiler, but there hasn't been a jansson release in a while, they have to add custom cflags or maybe even a patch just to shut up the warnings-turned-fatal-errors. They shouldn't have to.

Note that warnings can possibly help to find bugs, but they do not guarantee a bug. You added `void` in many places in this patch, and yet zero bugs were fixed. This still could help find and fix bugs due to calling functions with the wrong number of arguments, but there was no such bug thus far.

I am sorry if this comes across as argumentative ... I get the impression that many have come to consider "always use `-Werror`" a "modern best-practice" so I'm afraid it needs a lot of explaining why that is not really true.

coreyfarrell: I would be OK with our CI scripts inserting `-Werror` through CFLAGS but this cannot be done by default.  `-Werror` should only ever be used for development, adding it to releases would be a potability bug.

danielkempenich: Sure, within our organization, Werror is the best way to maintain a warning free build on all our platforms once it is warning free.  If that doesn't work for the jansson community I can remove it.  @ploxiln are you okay with the CI script approach as mentioned by @coreyfarrell?  I can make the changes when I get some free time, otherwise I'll just drop the Werror altogether.

ploxiln: Yes, `-Werror` makes sense for the CI build and test.

PR #520, opened 2020-01-09
Add support getn, setn functions

Mephistophiles:
Added support json_object_getn, setn, deln for non null-terminated strgins.

Please, review it. Thanks

### DIFF
--- a/doc/apiref.rst
+++ b/doc/apiref.rst
@@ -648,12 +648,28 @@ allowed in object keys.
    Get a value corresponding to *key* from *object*. Returns *NULL* if
    *key* is not found and on error.
 
+.. function:: json_t *json_object_getn(const json_t *object, const char *key, size_t key_len)
+
+   .. refcounting:: borrow
+
+   Like :func:`json_object_get`, but give the fixed-length *key* with length *key_len*.
+   See :ref:`fixed_length_keys` for details.
+
+   .. versionadded:: 2.14
+
 .. function:: int json_object_set(json_t *object, const char *key, json_t *value)
 
    Set the value of *key* to *value* in *object*. *key* must be a
    valid null terminated UTF-8 encoded Unicode string. If there
    already is a value for *key*, it is replaced by the new value.
    Returns 0 on success and -1 on error.
+
+.. function:: int json_object_setn(json_t *object, const char *key, size_t key_len, json_t *value)
+
+   Like :func:`json_object_set`, but give the fixed-length *key* with length *key_len*.
+   See :ref:`fixed_length_keys` for details.
+
+   .. versionadded:: 2.14
 
 .. function:: int json_object_set_nocheck(json_t *object, const char *key, json_t *value)
 
@@ -662,11 +678,25 @@ allowed in object keys.
    really is the case (e.g. you have already checked it by other
    means).
 
+.. function:: int json_object_setn_nocheck(json_t *object, const char *key, size_t key_len, json_t *value)
+
+   Like :func:`json_object_set_nocheck`, but give the fixed-length *key* with length *key_len*.
+   See :ref:`fixed_length_keys` for details.
+
+   .. versionadded:: 2.14
+
 .. function:: int json_object_set_new(json_t *object, const char *key, json_t *value)
 
    Like :func:`json_object_set()` but steals the reference to
    *value*. This is useful when *value* is newly created and not used
    after the call.
+
+.. function:: int json_object_setn_new(json_t *object, const char *key, size_t key_len, json_t *value)
+
+   Like :func:`json_object_set_new`, but give the fixed-length *key* with length *key_len*.
+   See :ref:`fixed_length_keys` for details.
+
+   .. versionadded:: 2.14
 
 .. function:: int json_object_set_new_nocheck(json_t *object, const char *key, json_t *value)
 
@@ -675,11 +705,25 @@ allowed in object keys.
    really is the case (e.g. you have already checked it by other
    means).
 
+.. function:: int json_object_setn_new_nocheck(json_t *object, const char *key, size_t key_len, json_t *value)
+
+   Like :func:`json_object_set_new_nocheck`, but give the fixed-length *key* with length *key_len*.
+   See :ref:`fixed_length_keys` for details.
+
+   .. versionadded:: 2.14
+
 .. function:: int json_object_del(json_t *object, const char *key)
 
    Delete *key* from *object* if it exists. Returns 0 on success, or
    -1 if *key* was not found. The reference count of the removed value
    is decremented.
+
+.. function:: int json_object_deln(json_t *object, const char *key, size_t key_len)
+
+   Like :func:`json_object_del`, but give the fixed-length *key* with length *key_len*.
+   See :ref:`fixed_length_keys` for details.
+
+   .. versionadded:: 2.14
 
 .. function:: int json_object_clear(json_t *object)
 
@@ -750,7 +794,7 @@ allowed in object keys.
    The items are returned in the order they were inserted to the
    object.
 
-   **Note:** It's not safe to call ``json_object_del(object, key)``
+   **Note:** It's not safe to call ``json_object_del(object, key)`` or ``json_object_deln(object, key, key_len)``
    during iteration. If you need to, use
    :func:`json_object_foreach_safe` instead.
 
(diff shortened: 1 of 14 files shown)


test/suites/api/test_object.c
```diff
@@ -780,6 +780,73 @@ static void test_bad_args(void) {
...
+        fail("json_object_getn on NULL failed");
+
+    if (!json_object_set_new(obj, NULL, json_true()))
+        fail("json_object_set_new with NULL key failed");
+
+    if (!json_object_set_new_nocheck(obj, NULL, json_true()))
+        fail("json_object_set_new_nocheck with NULL key failed");
+
+    if (!json_object_del(obj, NULL))
+        fail("json_object_del with NULL failed");
```
AllenX2018: the ```obj``` need to be released by ```json_decref(obj)```

src/jansson.h
```diff
@@ -186,11 +187,12 @@ static JSON_INLINE enum json_error_code json_error_code(const json_error_t *e) {
...
-json_t *json_object_get(const json_t *object, const char *key)
+json_t *json_object_getn(const json_t *object, const char *key, size_t key_len)
     JANSSON_ATTRS((warn_unused_result));
-int json_object_set_new(json_t *object, const char *key, json_t *value);
-int json_object_set_new_nocheck(json_t *object, const char *key, json_t *value);
-int json_object_del(json_t *object, const char *key);
+int json_object_setn_new(json_t *object, const char *key, size_t key_len, json_t *value);
+int json_object_setn_new_nocheck(json_t *object, const char *key, size_t key_len,
+                                 json_t *value);
+int json_object_deln(json_t *object, const char *key, size_t key_len);
```
AllenX2018: I'm afraid that it's not a good idea to remove the export api function

test/suites/api/test_object.c
Mephistophiles: Fixed, thanks

src/jansson.h
Mephistophiles: also thought about it, returned it back.

Mephistophiles:
@akheron, @coreyfarrell please comment this PR

doc/apiref.rst
```diff
@@ -681,6 +681,58 @@ allowed in object keys.
...
+
+   Get a value corresponding to *key* no more *key_len* chars from *object*. Returns *NULL* if
+   *key* is not found and on error.
+
+   .. versionadded:: 2.13
+
+.. function:: int json_object_setn(json_t *object, const char *key, size_t key_len, json_t *value)
+
+   Set the value of *key* to *value* in *object*. *key* must be a
+   valid null terminated UTF-8 encoded Unicode string. If there
```
ploxiln: The pull-request description says this is for non-null-terminated byte strings?

And above you say "*key* no more *key_len* chars", do you really mean "*key* exactly *key_len* chars"?

src/hashtable.c
```diff
@@ -79,7 +79,7 @@ static pair_t *hashtable_find_pair(hashtable_t *hashtable, bucket_t *bucket,
     list = bucket->first;
     while (1) {
         pair = list_to_pair(list);
-        if (pair->hash == hash && strcmp(pair->key, key) == 0)
+        if (pair->hash == hash && strncmp(pair->key, key, key_len) == 0)
```
ploxiln: `strncmp()` doesn't work exactly right for this: if `key` is a prefix of `pair->key` then this will think they match. You need to compare the length of the `pair->key` to `key_len` first. And if you want to support a `'\0'` in the middle of a key string (which `json_stringn()` supports), then you need to store the key length explicitly in every `pair_t`.

doc/apiref.rst
Mephistophiles: Fixed, thanks!

src/hashtable.c
Mephistophiles: Fixed, thanks

src/hashtable.c
```diff
@@ -79,7 +79,8 @@ static pair_t *hashtable_find_pair(hashtable_t *hashtable, bucket_t *bucket,
     list = bucket->first;
     while (1) {
         pair = list_to_pair(list);
-        if (pair->hash == hash && strcmp(pair->key, key) == 0)
+        if (pair->hash == hash && pair->key_len == key_len &&
+            strncmp(pair->key, key, key_len) == 0)
```
ploxiln: this `strncmp()` should be `memcmp()` in case there is a null in the middle (and because we already know the lengths match)

test/suites/api/test_object.c
```diff
@@ -780,6 +780,75 @@ static void test_bad_args(void) {
     json_decref(num);
 }
 
+static void test_keylen(void) {
+    json_t *obj = json_object();
+    const char key[] = {'t', 'e', 's', 't', '1'};
+    const char key2[] = {'t', 'e', 's', 't'};
```
ploxiln: you might want to add a test case with a key containing a null, like:

```c
    const char key3[] = {'t', 'e', 's', '\0', 't'};
```

src/hashtable.c
Mephistophiles: Replaced, thanks for your review!

test/suites/api/test_object.c
Mephistophiles: Added, plus added sample:
const char key4[] = {'t', 'e', 's', 't', '\0'};

Mephistophiles:
@akheron, @coreyfarrell friendly bump. Let's decide a destiny for this PR: either close, or rework, or merge this changes. Thanks!

akheron:
Hi, and sorry for not having a look earlier.

So, in essence, this PR adds support for object keys that are not null terminated AND may contain embedded null bytes. Am I right? In general, this sounds like a useful feature to me.

I'm a bit concerned about the extra `size_t` member of `hashtable_pair`, which makes hashtables grow 8 bytes times the number of entries on 64-bit machines. Does @coreyfarrell have an opinion on this?

It seems you null terminate the keys, which is important for iteration, good! `json_object_iter_key_len()` function should be added to get the key length at iterator.

The RFC conformance section of documentation should be updated, as it currently states that embedded null bytes are not supported in object keys. The API reference docs of the new functions should also be

coreyfarrell:
I suspect `src/dump.c` and `src/load.c` need some more work to fully deal with embedded nulls.  Can we add tests for dumping and loading objects such as `{"null\u0000embedded": true, "null": false}`?  At minimum we need to verify that round-tripping an object containing these two keys works.  Regarding json_dumps am I correct to assume `JSON_SORT_KEYS` will always sort `\u0000` first?  So a sorted object might be `{"\u0000": "first", "0": "second"}`?  Maybe test this to ensure that sorting happens to the raw key rather than the JSON encoded key (`\` comes after `0`)?  Also need to confirm that `JSON_SORT_KEYS` sees past an embedded null, so for example parsing `{"k\u0000-2": "second", "k\u0000-1": "first"}` then dumping sorted should reverse the keys.

> Hi, and sorry for not having a look earlier.

Same.  Unfortunately I've had other commitments lately and haven't had time to look at jansson lately.

> So, in essence, this PR adds support for object keys that are not null terminated AND may contain embedded null bytes. Am I right? In general, this sounds like a useful feature to me.
> 
> I'm a bit concerned about the extra `size_t` member of `hashtable_pair`, which makes hashtables grow 8 bytes times the number of entries on 64-bit machines. Does @coreyfarrell have an opinion on this?

JSON supports `{"embedded\u0000null": "value"}` so jansson probably should as well.  I'm a fan of minimizing storage but I think JSON compliance is worth it here.  Might even result in a slight CPU optimization by allowing internal code to avoid calls to strlen?

> It seems you null terminate the keys, which is important for iteration, good! `json_object_iter_key_len()` function should be added to get the key length at iterator.

Adding the null terminator is good/needed for backwards compatibility (preventing buffer over-reads) but it should be avoided by programs which use these new `setn` based functions or iterate object created by `json_load*`.  Otherwise the `"embedded\u0000null"` key would be incorrectly interpreted as just `"embedded"` (no buffer overrun but still a bug).

> The RFC conformance section of documentation should be updated, as it currently states that embedded null bytes are not supported in object keys. The API reference docs of the new functions should also be more clear about allowing embedded null bytes.

The version references throughout will probably need to be updated to say 2.14.

doc/apiref.rst
```diff
@@ -648,39 +648,87 @@ allowed in object keys.
...
+   .. versionadded:: 2.14
+
 .. function:: int json_object_set_nocheck(json_t *object, const char *key, json_t *value)
 
    Like :func:`json_object_set`, but doesn't check that *key* is
    valid UTF-8. Use this function only if you are certain that this
    really is the case (e.g. you have already checked it by other
    means).
 
+.. function:: int json_object_setn_nocheck(json_t *object, const char *key, json_t *value)
```
ploxiln, line 681: missing `key_len`

Mephistophiles: Oh, thanks! But I will add more documentation. Now there is no additional information about how to work with fixed-len keys.

Mephistophiles:
@akheron @coveralls  thanks for your response!
> json_object_iter_key_len() function should be added to get the key length at iterator.

Implemented. Thanks

> The RFC conformance section of documentation should be updated

Updated.

> The API reference docs of the new functions should also be more clear about allowing embedded null bytes.

Added the new section `Fixed-Length keys`, please review this

> Can we add tests for dumping and loading objects such as `{"null\u0000embedded": true, "null": false}`

Done.

>  Maybe test this to ensure that sorting happens to the raw key rather than the JSON encoded key (\ comes after 0)?

Done.

> Might even result in a slight CPU optimization by allowing internal code to avoid calls to strlen?

Done.

> The version references throughout will probably need to be updated to say 2.14.

Done

I also added 2 new iterators(json_object_keylen_foreach/json_object_keylen_foreach_safe), but I doubt if its are needed and what to call it.

akheron:
Merged, thanks! And sorry for the delay :(

Mephistophiles:
Thanks!

SamyDgz (approved):

PR #534, opened 2020-05-06
add jansson_* to shared library exports

xry111:
In 76300601d `jansson_version_str` and `jansson_version_cmp` were added.  But they are not in the shared library interface.  So with `--disable-static` in `configure` line, `make check` fails with `undefined reference to 'jansson_version_cmp'`.

This PR fixes it.


akheron:
Merged, thanks!

PR #535, opened 2020-05-06
add `scripts` directory to EXTRA_DIST

xry111:
Now jansson-2.13 tarball does not contain `scripts` directory, causing clang-format-check to fail.  This PR fixes that.


akheron:
Merged, thanks!

SamyDgz (approved):

PR #539, opened 2020-06-11
add json_object_getn for substring

irov:
Hi, I am writing a parser for text and I often need to replace tags whose table is in jansson. And it would be convenient for me to use a string segment.
like json_string_setn

example: "Hello, $[name]", i find $[ and ] and send to json_object_getn ptr for 'name]' and len equal 4

Thanks!


Mephistophiles:
Hi, The PR #520 has already implement similar functions.

irov:
> Hi, The PR #520 has already implement similar functions.

Ok. Thanks!

akheron:
Added in #520

PR #544, opened 2020-08-08
Lift "Depth of nested values" to a higher level

i-ky:
Currently this section is under "Numbers" which seems to be wrong.


akheron:
Merged, thanks!

SamyDgz (approved):

PR #554, opened 2020-11-05
Fix size_t related compiler warnings

kiyolee:
For Windows 64-bit build using MSVC, the fact that sizeof(long) != sizeof(size_t) would generate a fair bit compiler warnings.
Using size_t consistently is a better approach than using long and size_t interchangeably.
For printing size_t variables using printf(), It is also better to take advantage of C11 "%zd" if possible.


kiyolee:
should submit PR with branch other than master.

PR #555, opened 2020-11-05
Fix size_t related MSVC compiler warnings

kiyolee:
For Windows 64-bit build using MSVC, the fact that sizeof(long) != sizeof(size_t) would generate a fair bit compiler warnings.
Using size_t consistently is a better approach than using long and size_t interchangeably.
For printing size_t variables using printf(), It is also better to take advantage of C11 "%zd" if possible.

### DIFF
--- a/examples/simple_parse.c
+++ b/examples/simple_parse.c
@@ -90,11 +90,7 @@ void print_json_object(json_t *element, int indent) {
     print_json_indent(indent);
     size = json_object_size(element);
 
-#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
-    printf("JSON Object of %zd pair%s:\n", size, json_plural(size));
-#else
     printf("JSON Object of %lld pair%s:\n", (long long)size, json_plural(size));
-#endif
     json_object_foreach(element, key, value) {
         print_json_indent(indent + 2);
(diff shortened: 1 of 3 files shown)


examples/simple_parse.c
```diff
@@ -103,7 +107,11 @@ void print_json_array(json_t *element, int indent) {
     size_t size = json_array_size(element);
     print_json_indent(indent);
 
-    printf("JSON Array of %ld element%s:\n", size, json_plural(size));
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
+    printf("JSON Array of %zd element%s:\n", size, json_plural(size));
+#else
+    printf("JSON Array of %lld element%s:\n", (long long)size, json_plural(size));
```
ploxiln, line 113: If `%lld` and the `(long long)` cast works for all compilers, then I think it would be better to just do that, and omit the other 4 lines of code.

kiyolee: Fine with that. Let me fix it.

ploxiln: Thanks. The rest looks good to me, but I have no authority here :)

akheron:
Merged, thanks!

PR #557, opened 2020-11-10
Link to libm when necessary

RedDwarf69:
I have a system where isnan is a macro redirecting to __isnan(), which is only defined in libm.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -385,6 +385,19 @@ else()
       POSITION_INDEPENDENT_CODE true)
 endif()
 
+check_c_source_compiles("#include <math.h>
+                         int main() { return isnan(1.0); }" HAVE_ISNAN_WITHOUT_LIBM)
+if(NOT HAVE_ISNAN_WITHOUT_LIBM)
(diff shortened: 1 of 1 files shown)


akheron:
Would it be enough in your system to use `-ljansson -lm`?

RedDwarf69:
Yes, that would work.
But if you don't want the patch because it adds complexity for a corner case I will probably just carry the patch locally instead of adding the "fake" direct libm dependency to everything using jansson.

akheron:
Well, at least it would need to be ported to autoconf also

RedDwarf69:
I am doing as paid work and realistically I am not going to do it for autotools any time soon.
There is really a need, though? With or without an autotools fix this fixes something for people using CMake.
It's not like they generate the same stuff anyway:
- The CMake build generates a janssonConfig.cmake file to use by other cmake projects
- The autotools build generates a libtool file to use by other autotools projects

Autoconf is all but dead (https://lists.gnu.org/archive/html/autoconf/2021-01/msg00049.html), requiring fixing the autotools build system when fixing the cmake build system is only going to stop people from providing cmake fixes.

PR #560, opened 2020-12-14
GitHub Actions

akheron:
Move from Travis to GitHub actions. Keep AppVeyor to build on older Visual Studio versions, too.


ploxiln:
just for macOS+autotools, it appears you need to add a step like:

```
    run: brew install automake autoconf
```

also, it looks like the plain "build" step of the autotools job is missing, not sure if that's intentional

akheron:
There’s a lot still missing. My goal is to finish this in the near future :)

akheron:
I think everything else except ossfuzz is now included

akheron:
Decided to keep AppVeyor because GitHub actions only seems to support the newest Visual Studio.

SamyDgz (approved):

PR #564, opened 2021-01-19
Add json_strtok()

oliv3:
Hi,

First of all, many thanks for this wonderful library !
Would you be interested in a json_strtok() function ?
If so, I can add documentation and tests.

Best regards,

### DIFF
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -938,14 +938,13 @@ int json_unpack(json_t *root, const char *fmt, ...) {
 
 json_t *json_strtok(const char *str, const char *delim) {
     json_t *value = NULL;
-    char *s;
-    char *token;
+    char *s, *p, *token;
 
     if (str != NULL) {
         value = json_array();
 
         s = strdup(str);
-        for (token = strtok(s, delim); token; token = strtok(NULL, delim)) {
+        for (token = strtok_r(s, delim, &p); token; token = strtok_r(NULL, delim, &p)) {
             json_array_append_new(value, json_string(token));
         }
         free(s);
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -936,6 +936,10 @@ int json_unpack(json_t *root, const char *fmt, ...) {
     return ret;
 }
 
+#if defined(_WIN32)
+#define strtok_r strtok_s
+#endif
+
 json_t *json_strtok(const char *str, const char *delim) {
     json_t *value = NULL;
     char *s, *p, *token;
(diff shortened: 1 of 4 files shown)


src/pack_unpack.c
```diff
@@ -935,3 +935,21 @@ int json_unpack(json_t *root, const char *fmt, ...) {
...
+json_t *json_strtok(const char *str, const char *delim) {
+    json_t *value = NULL;
+    char *s;
+    char *token;
+
+    if (str != NULL) {
+        value = json_array();
+
+        s = strdup(str);
+        for (token = strtok(s, delim); token; token = strtok(NULL, delim)) {
```
ploxiln, line 948: This should use `strtok_r()` to be safe in a multi-threaded program.

However, I think `json_strtok()` is probably out of scope for this library. (But I'm not a maintainer :)

oliv3: > This should use `strtok_r()` to be safe in a multi-threaded program.

Sure.

> However, I think `json_strtok()` is probably out of scope for this library. (But I'm not a maintainer :)

I dont't know. I needed it, maybe others do too ?

oliv3:
@ploxiln Any idea why the `fuzz` test fails ?

ploxiln:
That's a mis-configuration of the fuzz test, not your fault. (Your branch name doesn't exist on this remote, but there's a different way you can refer to a numbered virtual branch github makes for each PR...)

```
Step 2/4 : RUN git clone --depth 1 -b json_strtok https://github.com/akheron/jansson.git /src/jansson
 ---> Running in c1d3bf576168
Cloning into '/src/jansson'...
warning: Could not find remote branch json_strtok to clone.
fatal: Remote branch json_strtok not found in upstream origin
```

src/pack_unpack.c
coreyfarrell: Since the purpose of your function is to parse text which is not JSON I would agree that this is out of scope.  That is not a judgement on the potential usefulness, just that jansson is a specialized library rather than a generic library so I think it should not parse non-JSON text.

In addition based on the test you submitted I should point out that `strtok` is not a proper CSV parser which would require processing quoted strings.

oliv3: > In addition based on the test you submitted I should point out that `strtok` is not a proper CSV parser which would require processing quoted strings.

Yeah using CSV was just for the test, it does what `strotk_r()` does.

Feel free to close the MR when you all agree it is out of scope. Thanks for jansson anyway !

akheron:
Hi, and thanks for the PR. Unfortunately I’m with @ploxiln here and think this is out of scope of Jansson.

PR #571, opened 2021-01-27
cmake: honour CMAKE_INSTALL_* variables

RedDwarf69:
It ignores the previous JANSSON_INSTALL_*_DIR variables. But they were
broken, only affecting the pkg-config creation logic, not the actual
installation directory. This can't break anybody's existing usage.


JosiahWI:
Is it possible this would fix #668?

PR #573, opened 2021-02-16
fix #572 

### LINKED ISSUE #572: LLVM/LLD error: unknown argument '--default-symver'
Looks like this commit (https://github.com/akheron/jansson/commit/ca80d5127e56dd2ee056fdf3f4204017b78b89e0) does not takes into account that defult LD may be set to LLVM/LLD, finally producing error during linking:

```
FAILED: lib/libjansson.so.4.13.0 
: && /usr/bin/clang -fPIC -Os -fomit-frame-pointer -g1 -Wstrict-aliasing=2 -pipe -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fstack-protector-all --param=ssp-buffer-size=4 -m64 -mtune=generic -flto -Os -fomit-frame-pointer -g1 -Wstrict-aliasing=2 -pipe -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fstack-protector-all --param=ssp-buffer-size=4 -m64 -mtune=generic -flto  -Os -fomit-frame-pointer -g1 -Wstrict-aliasing=2 -pipe -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fstack-protector-all --param=ssp-buffer-size=4 -m64 -mtune=generic -flto -Wl,-O2  -Wl,--no-undefined -flto -Wl,--default-symver -shared -Wl,-soname,libjansson.so.4 -o lib/libjansson.so.4.13.0 CMakeFiles/jansson.dir/src/dump.c.o CMakeFiles/jansson.dir/src/error.c.o CMakeFiles/jansson.dir/src/hashtable.c.o CMakeFiles/jansson.dir/src/hashtable_seed.c.o CMakeFiles/jansson.dir/src/load.c.o CMakeFiles/jansson.dir/src/memory.c.o CMakeFiles/jansson.dir/src/pack_unpack.c.o CMakeFiles/jansson.dir/src/strbuffer.c.o CMakeFiles/jansson.dir/src/strconv.c.o CMakeFiles/jansson.dir/src/utf.c.o CMakeFiles/jansson.dir/src/value.c.o CMakeFiles/jansson.dir/src/version.c.o   && :
ld: error: unknown argument '--default-symver'
clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
```

P.s.
Looks like only ld.bfd does support this https://bugs.llvm.org/show_bug.cgi?id=43452

tpgxyz:
use --version-script in case of linkers that does not support --default-symver
This is WIP to fix #572

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,9 +57,6 @@ if (HAS_NO_FORMAT_TRUNCATION)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-format-truncation")
 endif()
 
-include (CheckSymbolExists)
-check_symbol_exists("__GLIBC__" "stdlib.h" LIBC_IS_GLIBC)
-
 if (MSVC)
    # Turn off Microsofts "security" warnings.
    add_definitions( "/W3 /D_CRT_SECURE_NO_WARNINGS /wd4005 /wd4996 /nologo" )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -311,8 +314,17 @@ if(JANSSON_BUILD_SHARED_LIBS)
       ${JANSSON_HDR_PUBLIC}
       src/jansson.def)
 
-   if (LIBC_IS_GLIBC)
+   if (LIBC_IS_GLIBC AND LD_SYMVER)
       set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--default-symver")
+   else()
+      file(WRITE "${CMAKE_SOURCE_DIR}/jansson.map" "libjansson.so.${JANSSON_SOVERSION} {
```
ploxiln, line 320: Any chance this can go in the build dir instead, to avoid writing files to the source dir? I'm not experienced with cmake, but it looks like that might be `${CMAKE_CURRENT_BINARY_DIR}`.

Also, this could probably just switch to `--version-script` unconditionally and it would work for all linkers.

tpgxyz: Sure, will update that.

tpgxyz: Ok done

tpgxyz:
Looks like your CI test does use LLVM/lld as linker.
I tested this locally with CC=clang and LD=lld and when i run` llvm-readobj --dyn-syms *.so` it list correct list of symbols.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.1)
+cmake_minimum_required (VERSION 3.18.6)
```
ploxiln, line 1: the appveyor CI (for windows) has cmake 3.16.2, so this is why those fail ... and we probably don't want to raise the required cmake version unnecessarily, for example ubuntu 16.04 has cmake 3.5.1

tpgxyz: `check_linker_flag` was introduced in cmake 3.18 in may 2020.

Well I think that distros should keep cmake updated.

You want to update your CI or want me to provide some primitives as replacement for that cmake module?

ploxiln: Well, it's not my project.

But, this seems a good argument to just use the linker script which works with all linkers, so this CMakeLists will remain compatible with linux distros and embedded build toolchains which most people still use.

tpgxyz: Fixed

tpgxyz:
Ok i've make it more simplified.
First cmake check for --default-symver and if test fails it checks for --version-script

tpgxyz:
@ploxiln ping

akheron:
Merged, thanks!

tpgxyz:
@akheron Thanks.

PR #582, opened 2021-06-21
jsonp_strtod: return -1 on Infinity or NaN

guijan:
strtod() also looks for the strings "INF" or "NAN" and returns Infinity or NaN respectively, both are explicitly forbidden by the RFC.

### DIFF
--- a/src/strconv.c
+++ b/src/strconv.c
@@ -71,6 +71,12 @@ int jsonp_strtod(strbuffer_t *strbuffer, double *out) {
         /* Overflow */
         return -1;
     }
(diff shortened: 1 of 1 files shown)


akheron:
Thanks! Would you mind adding a test that demonstrates the case this change fixes? Also, it seems that this doesn’t compile on older Visual Studio versions, which has to be worked around somehow.

guijan:
I'm sorry for the trouble, but I only realized now the parser detects a syntax error in the JSON before strtod gets to convert such a string.

PR #610, opened 2022-04-28
Fix overwriting linker flags

Thomas1664:
You were overwriting `CMAKE_SHARED_LINKER_FLAGS` without caring about already set flags. This causes builds on UWP to fail.


akheron:
Merged, thanks!

PR #620, opened 2022-07-09
Clean up header files somewhat

eriols:


eriols:
I was running include_what_you_use on a library at my day job that has jansson as a submodule and noticed these suggestions, so I applied the ones that made sense to me here.

PR #622, opened 2022-07-21
Fix the read callback, function prototype

dittonedo45:
Set the json_load_callback_t prototype, the usual way liket that of read(int, void*, size_t);


src/jansson.h
```diff
@@ -375,6 +375,8 @@ json_t *json_load_file(const char *path, size_t flags, json_error_t *error)
     JANSSON_ATTRS((warn_unused_result));
 json_t *json_load_callback(json_load_callback_t callback, void *data, size_t flags,
                            json_error_t *error) JANSSON_ATTRS((warn_unused_result));
+json_t *json_load_callback1(json_load_callback_t1 callback, void *data, size_t flags,
```
Mephistophiles, line 378: What is json_load_callback1?

PR #662, opened 2023-09-28
A new attempt at location information for decoded JSON elements

n0-1:
This is a respin of the rejected #461 containing a reimplementation of the
feature in a less intrusive manner for library users keeping it disabled.

The nftables project is very interested in having this feature to provide
similar error reporting to its JSON API users as there is for non-JSON input.

This new implementation does not introduce any overhead on storage and
negligible on CPU cycles if not enabled: It then boils down to a few extra
function calls returning early after a simple variable value check.

The above is realized by storing json_t objects' location data in a hash table
indexed by the object's address. The table is initialized upon first insert,
which then registers a destructor function using atexit() to free up any data
at program exit (to avoid complaints from memory checkers like valgrind).

If an object is deleted, its corresponding hashtable entry is removed along
with it. To facilitate location data for simple values, the singleton 'true',
'false' and 'null' objects have to be turned into allocated data structures (so
they get a unique address in memory).

This MR contains five commits:

1. is fallout, a previous commit missed to update tests/.gitignore.
2. adds a function to print a format string with negative test results.
3. introduces the new decoder flag.
4. implements refcounted simple values.
5. contains the actual implementation.


n0-1:
Rebased onto current HEAD, no functional changes intended.

This merge request is lingering uncommented for over a year now. Quite frustrating to receive no feedback at all.

Is the implemented feature really unnecessary? How do other folks report semantic errors in JSON input?

Zer0-One:
It seems like a cool feature to me, and I appreciate that it's optional. @akheron thoughts?

PR #664, opened 2023-09-28
Use target-based cmake settings

Andrew-Au:
- Update minimum required to CMake version 3.5 (versions older than 3.5 are deprecated as of 3.27)
- update add_definitions to target_compile_definitions
- use target_include_directories for public library includes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.1)
+cmake_minimum_required (VERSION 3.5)
 project(jansson C)
 
 # Options
@@ -271,17 +271,14 @@ configure_file (${CMAKE_CURRENT_SOURCE_DIR}/cmake/jansson_config.h.cmake
 file (COPY ${CMAKE_CURRENT_SOURCE_DIR}/src/jansson.h
            DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/include/)
 
-add_definitions(-DJANSSON_USING_CMAKE)
-
 # configure the private config file
 configure_file (${CMAKE_CURRENT_SOURCE_DIR}/cmake/jansson_private_config.h.cmake
                 ${CMAKE_CURRENT_BINARY_DIR}/private_include/jansson_private_config.h)
 
-# and tell the source code to include it
-add_definitions(-DHAVE_CONFIG_H)
-
-include_directories (${CMAKE_CURRENT_BINARY_DIR}/include)
 include_directories (${CMAKE_CURRENT_BINARY_DIR}/private_include)
+
+# Configuration flags will be set on project later once we have defined the target
+
 
 # Add the lib sources.
 file(GLOB JANSSON_SRC src/*.c)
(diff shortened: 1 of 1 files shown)


akheron:
Thanks! I'm not an expert in CMake, but if experts say that this is the right way to do it, I'm inclined to believe 😄 

Can you clarify why the `Jansson::jansson` alias is needed?

CMakeLists.txt
```diff
@@ -358,6 +355,20 @@ else()
       POSITION_INDEPENDENT_CODE true)
 endif()
 
+
+# Now target jansson is declared, set per-target values
+
+target_compile_definitions(jansson PUBLIC JANSSON_USING_CMAKE)
+target_compile_definitions(jansson PRIVATE HAVE_CONFIG_H)
```
JosiahWI, line 362:
```suggestion
target_compile_definitions(jansson PUBLIC JANSSON_USING_CMAKE PRIVATE HAVE_CONFIG_H)
```

Might be good to split across multiple lines for clarity, but there's no need to use two separate invocations of the `target_compile_definitions` command here.

CMakeLists.txt
```diff
@@ -358,6 +355,20 @@ else()
...
+
+target_include_directories(jansson PUBLIC
+	$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
+	INTERFACE $<INSTALL_INTERFACE:include>
+)
+
+add_library( Jansson::jansson ALIAS jansson )
```
JosiahWI, line 369: This change is unrelated to the topic of this PR. An alias is often added for consistency so that the target name is the same whether you use `add_subdirectory`, `FetchContent`, or `find_package`.

CMakeLists.txt
```diff
@@ -358,6 +355,20 @@ else()
...
+
+# Now target jansson is declared, set per-target values
+
+target_compile_definitions(jansson PUBLIC JANSSON_USING_CMAKE)
+target_compile_definitions(jansson PRIVATE HAVE_CONFIG_H)
+
+target_include_directories(jansson PUBLIC
+	$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
+	INTERFACE $<INSTALL_INTERFACE:include>
+)
```
JosiahWI, line 367: Nice, looks good. The formatting seems a bit inconsistent with `PUBLIC` on the line before its argument and `INTERFACE` on the same line as its argument.

CMakeLists.txt
```diff
@@ -271,18 +271,15 @@ configure_file (${CMAKE_CURRENT_SOURCE_DIR}/cmake/jansson_config.h.cmake
...
 configure_file (${CMAKE_CURRENT_SOURCE_DIR}/cmake/jansson_private_config.h.cmake
                 ${CMAKE_CURRENT_BINARY_DIR}/private_include/jansson_private_config.h)
 
-# and tell the source code to include it
-add_definitions(-DHAVE_CONFIG_H)
-
-include_directories (${CMAKE_CURRENT_BINARY_DIR}/include)
 include_directories (${CMAKE_CURRENT_BINARY_DIR}/private_include)
 
+# Configuration flags will be set on project later once we have defined the target
```
JosiahWI, line 280: Thanks for adding comments.

CMakeLists.txt
JosiahWI: It should be `jansson::jansson` with a lowercase j, to be consistent with the exported namespace.

CMakeLists.txt
```diff
@@ -271,18 +271,15 @@ configure_file (${CMAKE_CURRENT_SOURCE_DIR}/cmake/jansson_config.h.cmake
...
-
 # configure the private config file
 configure_file (${CMAKE_CURRENT_SOURCE_DIR}/cmake/jansson_private_config.h.cmake
                 ${CMAKE_CURRENT_BINARY_DIR}/private_include/jansson_private_config.h)
 
-# and tell the source code to include it
-add_definitions(-DHAVE_CONFIG_H)
-
-include_directories (${CMAKE_CURRENT_BINARY_DIR}/include)
 include_directories (${CMAKE_CURRENT_BINARY_DIR}/private_include)
```
JosiahWI, line 278: Why isn't this one moved as well?

PR #666, opened 2023-10-12
Port check for --default-symver to autoconf

vimproved:
This commit ports the configure check for -Wl,--default-symver that is present in CMake to autoconf. This fixes building Jansson via autoconf with non-bfd linkers on glibc systems.


thesamesam:
This PR was missed (like my other one unfortunately, fixing a different but related problem) but looks like this was fixed by https://github.com/akheron/jansson/commit/272d75a5e1685fc42f7123731f629e13af1942c0.

PR #639, opened 2022-11-21
CMake: fix typo in default-symver check

thesamesam:
Signed-off-by: Sam James <sam@gentoo.org>

PR #681, opened 2024-03-18
fix --default-symver can not be set when use cmake

zhangwenlong8911:


Zer0-One:
May also be an option to use https://cmake.org/cmake/help/latest/module/CheckLinkerFlag.html

JosiahWI:
This is an alternative solution to #639. Note that CMake variables are not usually named with a D prefix like C defines are. The Find modules written by Kitware are good examples.

thesamesam:
Handled by https://github.com/akheron/jansson/pull/750 now I think.

PR #684, opened 2024-03-22
Implement json_compare() function.

solemnwarning:
This commit introduces a `json_compare()` function which can be used to compare `json_t structures for sorting purposes.


solemnwarning:
I'm not really sure why the CI is failing for VS <2013.

> 1>..\src\value.c(746) : error C2143: syntax error : missing ';' before 'type'

```c
    for (i = 0; i < min_size; i++) {
        json_t *value1, *value2;

        value1 = json_array_get(array1, i);
        value2 = json_array_get(array2, i);

        int result = json_compare(value1, value2); /* <-- Line 746 */
        if (result != 0)
            return result;
    }
```

solemnwarning:
That Ubuntu CI job highlighted the problem... been a long time since I've had to write C89 compliant code.

My use case here is a map which includes a blob of arbitrary JSON as part of the key. I've actually got it implemented by serialising it to a string and then comparing that as described, however I'm not a fan of it since it means either keeping the serialised form around, or serialising on every `operator<` call and potentially having an allocation failure.

For my use case at least, the data structures are small so the inefficient sorting doesn't really matter.

PR #688, opened 2024-04-17
Create Variadic Object Search

ChiragH25:
Searches Object using a variadic function.

Function takes a root object, a depth of search and a path composed of json_type types and char* keys for elements in objects and size_t indexs for objects in lists.

Note, getting an object from a list counts towards depth

examples below from tests, and reference the following JSON:
```
{
  "base": {
    "array": [{}, [], "string 1", 1, 2.2000000000000002, true, false, null],
    "string2": "string3",
    "integer": 3,
    "real": 4.4000000000000004,
    "true": true,
    "false": false,
    "null": null,
    "nest 1": { "nest 2": { "nest 3": {} } }
  }
}
```

`value = json_object_get_path(json, 2, JSON_OBJECT, "base", JSON_ARRAY, "array");`
returns a json object for
`[{}, [], "string 1", 1, 2.2000000000000002, true, false, null]`

`value = json_object_get_path(json, 3, JSON_OBJECT, "base", JSON_ARRAY, "array", (size_t) 2);`
returns a json object for
`"string 1"`

`value = json_object_get_path(json, 4, JSON_OBJECT, "base", JSON_OBJECT, "nest 1", JSON_OBJECT, "nest 2");`
returns a json object for
`{"nest 3":{}}`

### DIFF
--- a/src/value.c
+++ b/src/value.c
@@ -108,6 +108,13 @@ json_t *json_object_get_path(json_t* object, size_t depth, ...){
     va_start(path, depth);
     for (size_t i = 0; i < depth; i++)
     {
+        if(object->type == JSON_ARRAY){
+            object = json_array_get(object, va_arg(path, size_t));
+            if(!object){
+                return NULL;
+            }
+            continue;
+        }
         json_type object_t = va_arg(path, json_type);
         object = json_object_get(object, va_arg(path, const char*));
         if(!object){
@@ -115,15 +122,6 @@ json_t *json_object_get_path(json_t* object, size_t depth, ...){
         }
         if(object->type != object_t){
             return NULL;
-        }
-        if(object_t == JSON_ARRAY){
-            i++;
-            if(i<depth){
-                object = json_array_get(object, va_arg(path, size_t));
-                if(!object){
-                    return NULL;
-                }
-            }
         }
     }
 
--- a/src/value.c
+++ b/src/value.c
@@ -103,6 +103,33 @@ json_t *json_object_get(const json_t *json, const char *key) {
     return json_object_getn(json, key, strlen(key));
 }
 
+json_t *json_object_get_path(json_t* object, size_t depth, ...){
+    va_list path;
+    va_start(path, depth);
+    for (size_t i = 0; i < depth; i++)
+    {
+        json_type object_t = va_arg(path, json_type);
+        object = json_object_get(object, va_arg(path, const char*));
+        if(!object){
+            return NULL;
+        }
+        if(object->type != object_t){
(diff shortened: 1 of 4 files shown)


ChiragH25:
So, again sorry about the appveyor thing. it's 3AM here and I can't sleep so I though I'd fix this but I didn't realise I was pushing to the PR instead of my branch.

src/value.c
```diff
@@ -103,6 +103,31 @@ json_t *json_object_get(const json_t *json, const char *key) {
...
+
+    for (i = 0; i < depth; i++){
+        if(object->type == JSON_ARRAY){
+            object = json_array_get(object, va_arg(path, size_t));
+            if(!object){
+                return NULL;
+            }
+            continue;
+        }
+        object = json_object_get(object, va_arg(path, const char*));
```
Mephistophiles, line 120: Hi, thanks for the PR!

It seems to me that the function is quite dangerous: if the json structure differs from the expected one, then the passed array index can be used as a pointer to a string. 

In this case, we have no way to return an error to the user, since va args is not typed in any way.

ChiragH25: That is true, though that seems inherent to a variadic approach. printf and other functions I've found tend to have this issue regardless. I considered an approach of using a pointer array to a specific struct with appropriate typechecking but it defeated the purpose of making a convenient interface.

I largely created this function as a way of simplifying getting parts of a JSON object from a known JSON structure and thought it might be useful for other people as well. the variadic approach handles object traversal well and in my exeprience using this in my own project, I've found that if I ever mess up the location of an index in the path, I just get a null return since it's extremely unlikely that such a space in memory would have a char pointer with a value that's also a key in the object that is being queried.

And even if it did, the worst case result is that it returns an unintended child of the JSON which is still relatively safe.

I've also passed a char* as an index and just had null returns since it's unlikely that the array in question is large enough that the pointer in question is a valid index.

Since any inputs to this function are never edited or copied, simply read from the impact on other parts of the memory should be effectively null.

Granted this is obviously not intended for any application where the path is passed in by a user or some outside source, since that might allow a bad actor to pass in something specifically malicious. (though it would be quite unweildy to work with regardless in this scenario as far as I'm aware)

But in the bounds of standard development where this may be used, I figure that it fails in a sufficiently safe manner even if the  path is incorrect.

Beyond that if this still seems too unsafe, I'm not too concerned about it and just wanted to pass along a tool that I made since I thought it might be useful, and I'll just retain it for my own use. Thank you for your time in that case, and if you have any reccomendations on how to make it safer given the goal of this function, please let me know as I'm more than happy to learn.

Issue #630, opened 2022-09-01
compiler warning in call to strncpy


cosinekitty:
Using gcc 10.2.1 on a Debian Bullseye system, I encounter the following warning:

```
In function ‘jsonp_error_set_source’,
    inlined from ‘jsonp_error_set_source’ at error.c:17:6:
error.c:25:9: warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
   25 |         strncpy(error->source, source, length + 1);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
error.c: In function ‘jsonp_error_set_source’:
error.c:23:14: note: length computed here
   23 |     length = strlen(source);
      |              ^~~~~~~~~~~~~~
```

I doubt it is anything to worry about, but I thought you might like to know. The complete log of how I did this build is below, in case that helps.

Thanks for creating and maintaining this awesome library!

```
don@doctorno:~/github/jansson$ uname -a
Linux doctorno 5.10.0-17-amd64 #1 SMP Debian 5.10.136-1 (2022-08-13) x86_64 GNU/Linux
don@doctorno:~/github/jansson$ gcc --version
gcc (Debian 10.2.1-6) 10.2.1 20210110
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

don@doctorno:~/github/jansson$ git pull
Already up to date.
don@doctorno:~/github/jansson$ autoreconf -i
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, '.'.
libtoolize: copying file './ltmain.sh'
libtoolize: Consider adding 'AC_CONFIG_MACRO_DIRS([m4])' to configure.ac,
libtoolize: and rerunning libtoolize and aclocal.
libtoolize: Consider adding '-I m4' to ACLOCAL_AMFLAGS in Makefile.am.
configure.ac:11: installing './compile'
configure.ac:13: installing './config.guess'
configure.ac:13: installing './config.sub'
configure.ac:5: installing './install-sh'
configure.ac:5: installing './missing'
parallel-tests: installing './test-driver'
src/Makefile.am: installing './depcomp'
don@doctorno:~/github/jansson$ ./configure
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
[...]

PR #700, opened 2024-09-26
Compiler warning -Werror=stringop-truncation fix for error.c

sbytnar:
Hi. This warning is emitted by newer versions of gcc. The project I'm working with treats this particular warning as an error.
The fix below makes the compiler happy.

The full error message emitted looks like this:
```
| jansson/src/error.c: In function 'jsonp_error_set_source':
| jansson/src/error.c:25:9: error: '__builtin_strncpy' specified bound depends on the length of the source argument [-Werror=stringop-truncation]
|    25 |         strncpy(error->source, source, length + 1);
|       |         ^
| In function 'jsonp_error_set_source',
| jansson/src/error.c:17:6:
| jansson/src/error.c:23:14: note: length computed here
|    23 |     length = strlen(source);
|       |              ^~~~~~~~~~~~~~
| cc1: all warnings being treated as errors
```

Thank you.
--Steve

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -22,7 +22,7 @@ void jsonp_error_set_source(json_error_t *error, const char *source) {
 
     length = strlen(source);
     if (length < JSON_ERROR_SOURCE_LENGTH)
-        strncpy(error->source, source, length + 1);
+        strncpy(error->source, source, JSON_ERROR_SOURCE_LENGTH);
     else {
         size_t extra = length - JSON_ERROR_SOURCE_LENGTH + 4;
         memcpy(error->source, "...", 3);


sbytnar:
This would fix https://github.com/akheron/jansson/issues/630

SamyDgz (approved):

src/error.c
```diff
@@ -22,7 +22,7 @@ void jsonp_error_set_source(json_error_t *error, const char *source) {
 
     length = strlen(source);
     if (length < JSON_ERROR_SOURCE_LENGTH)
-        strncpy(error->source, source, length + 1);
+        strncpy(error->source, source, JSON_ERROR_SOURCE_LENGTH);
```
rosslagerwall, line 25: Passing `JSON_ERROR_SOURCE_LENGTH` here seems unnecessary as the length has just been checked. We already know the length of `source` and have checked it will fit in the output buffer, therefore I think this is better:

```
         memcpy(error->source, source, length + 1);
```

TroyHernandez:
Adding a current downstream reproduction: bundling Jansson 2.15.1 in an R package and compiling it with GCC 14.3.0 from Rtools45 (mingw-w64/UCRT) emits:

```
error.c:25:9: warning: 'strncpy' specified bound depends on the length of the source argument [-Wstringop-truncation]
```

`R CMD check` classifies this as a significant compilation warning and flags the package.

The second `strncpy` in `jsonp_error_set_source`, in the long-source branch, also triggers this warning. Both copies already have exact, checked lengths:

- when `length < JSON_ERROR_SOURCE_LENGTH`, `length + 1` includes the NUL and fits the destination;
- in the truncating branch, `length - extra + 1` equals the remaining destination capacity.

Replacing both calls with `memcpy` therefore preserves the existing behavior and silences the warning, matching the review suggestion above. We carry and test that patch in [cornball-ai/janssonr](https://github.com/cornball-ai/janssonr/blob/main/src/jansson/PATCHES.md). I'd be happy to open a superseding PR if this one is dormant.

fallenmi (approved):
Verified on the exact current base/head pair 61fc3d0e28e1a35410af42e329cd977095ec32d2 / 0e3b06d95a96775b70e6749927bf6e83cb3224e9. With GCC 16.1.0 and `-O2 -Wall -Wextra -Werror=stringop-truncation`, the base fails at `src/error.c:25`, while the exact merged head builds cleanly; all 213 CTest cases pass. The constant bound is safe under the preceding `length < JSON_ERROR_SOURCE_LENGTH` guard: `strncpy` copies the terminator and only pads the remaining destination bytes.

Disclosure: this review was prepared with Codex assistance; I independently verified the diff, the base/head oracle, and the test result.

PR #706, opened 2025-01-30
Add JANSSON_INSTALL_TESTS CMake Option

dchordia:
Addition of JANSSON_INSTALL_TESTS option installs tests at /bin/jansson_tests. It will be a nice add-on, keeping all the tests and data required for testing under single directory.


dchordia:
@akheron what do you think?

akheron:
Sorry, but I don’t see the point. When using CMake, tests are meant to be run with checked out source code using ctest. 

Furthermore, adding a subdirectory under /bin is nonstandard, at least in Unix-like systems.

PR #713, opened 2025-04-03
update cmake minimum required version to fix cmake 4.0.0 error

XielQs:


JosiahWI:
#664 bumped the minimum version to CMake 3.10.

hnyman:
Could there be a new release to get the already implemented cmake fix aef13f8 into a release?

PR #717, opened 2025-06-14
Add realloc

chromhelm:
Make use of realloc. This can be a significant performance improvement.

### DIFF
--- a/doc/apiref.rst
+++ b/doc/apiref.rst
@@ -1907,7 +1907,7 @@ Copying objects preserves the insertion order of keys.
 Custom Memory Allocation
 ========================
 
-By default, Jansson uses :func:`malloc()` and :func:`free()` for
+By default, Jansson uses :func:`malloc()`, :func:`realloc()` and :func:`free()` for
 memory allocation. These functions can be overridden if custom
 behavior is needed.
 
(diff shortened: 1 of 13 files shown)


doc/apiref.rst
```diff
@@ -1918,6 +1918,16 @@ behavior is needed.
 
        typedef void *(*json_malloc_t)(size_t);
 
+.. type:: json_realloc_t
+
+   A typedef for a function pointer with :func:`realloc()`'s
+   signature::
+
+       typedef void *(*json_malloc_t)(void*, size_t);
```
akheron, line 1926: Should be `json_realloc_t`

akheron:
Thanks for the PR!

I haven't benchmarked myself, but a quick google search suggests that malloc + free is faster when you don't want to keep the previous data.  `hashtable_do_rehash` seems to do just this, so maybe using realloc is not justified there.

doc/apiref.rst
WilhelmWiens: Yes of course. thanks.

WilhelmWiens:
The old implementation was optimized to both insertions in the middle and at the end.

With Realloc, appending is now faster, when the memory area could be expanded, and about as fast when the array had to be copied.
When inserting in the middle, the best case is about the same. At worst case, the memory after the insertion is copied twice now.

What do you think about the change in the array? @akheron

akheron:
Well, I've always maintained the guideline that jansson is not a data structure or algorithm library. As such, raw speed is not a priority, only a convenience 😄.

I think the array code is now better, as it's easier to understand AND faster in some cases.

PR #731, opened 2026-03-25
Unpack: include keys in type mismatch error messages

cwalther:
I think the error messages like `Expected string, got integer` that `json_unpack_ex` outputs when parsing a JSON object like `{"one": 1, "two": 2, "three": 3}` would be much more helpful if they included information about _which_ item it was that was found problematic – just like the error message for a missing item already does: `Object item not found: four`.

The attached commit is a proposal how this could be implemented: it now says `Expected string, got integer for key two`. For values that are not in an object, but in an array or at the root level, the error message is unmodified as before.

This is not a full path in the case of nested objects, just the key in the innermost object, however I think that will be sufficient to guide a user in most cases. In cases where it is not, the calling code can still determine the full path by examining `error.position` (which points at the corresponding format specifier in the format string).

If changing error messages for existing callers is considered an incompatible API change, then this functionality could also be gated behind an additional flag for `json_unpack_ex`.

What do you think?

### DIFF
--- a/src/pack_unpack.c
+++ b/src/pack_unpack.c
@@ -460,9 +460,10 @@ static json_t *pack(scanner_t *s, va_list *ap) {
     }
 }
 
-static int unpack(scanner_t *s, json_t *root, va_list *ap);
+static int unpack(scanner_t *s, json_t *root, va_list *ap, const char *key);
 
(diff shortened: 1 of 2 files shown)


cwalther:
Thanks!

PR #733, opened 2026-04-17
Machine-readable license information with REUSE

Volker-Weissmann:
Hello there,
[REUSE](https://reuse.software) is a standard that allows a project to specifiy license and copyright information in a machine-readable format (e.g. `reuse lint --json` now tells you which file has which license).

E.g. the ArchLinux project [also likes REUSE](https://rfc.archlinux.page/0052-reuse/).

The [specification says](https://reuse.software/spec-3.3/#comment-headers) that we SHOULD prefer using comment headers over specifying the licenses in `REUSE.toml`, but I don't know if the jansson maintainers are fine with me touching so many files.

If you are wondering why it says `Copyright (c) <year> <copyright holders>` in LICENSES/MIT.txt: According to [the REUSE FAQ](https://reuse.software/faq/#license-templates) it's what you are supposed to do, and e.g. [this project](https://github.com/nextcloud/files_pdfviewer/blob/master/LICENSES/MIT.txt) and [this project](https://github.com/organicmaps/organicmaps/blob/master/LICENSES/MIT.txt) did it the same.


Volker-Weissmann:
I could also insert `//SPDX-License-Identifier: ...` and `//SPDX-FileCopyrightText: ...` comments into the source files instead of using a REUSE.toml file. Both the REUSE specification and me would actually even prefer this, since in the future people might forgot to need to update REUSE.toml when they e.g. move a file. The disadvantage is that we would have a commit that touches a lot of files.

PR #734, opened 2026-04-26
use bits for type to improve type checking performance

zx80:
When checking JSON structures the type macros are used and can represent a significant execution time, especially if there is an implicit branch (`json_is_number` and `json_is_boolean`).

This patch uses bits instead of integer values to remove one branch and improve performance.

### DIFF
--- a/src/jansson.h
+++ b/src/jansson.h
@@ -48,14 +48,14 @@ extern "C" {
 /* types */
 
 typedef enum {
-    JSON_OBJECT,
-    JSON_ARRAY,
-    JSON_STRING,
-    JSON_INTEGER,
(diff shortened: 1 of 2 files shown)


akheron:
This change would break the ABI because the enum is public. Such a small improvement doesn’t warrant a new major version.

zx80:
Too bad, the performance improvement is quite significant on some benchmarks.

Your point suggests that if there is a new version which break the ABI at some point in the future, such a change could be considered?

akheron:
Well maybe, but given that the library is old and quite stable, I wouldn’t hold my breath.

If raw speed is a priority for you, some other json library might be a better fit.

PR #744, opened 2026-06-24
configure: only use --default-symver when the linker supports it

kraj:
jansson enables symbol versioning on glibc by passing -Wl,--default-symver to the linker. This is a GNU ld option which is not understood by every linker, e.g. ld.lld errors out with:

  ld.lld: error: unknown argument '--default-symver'

Probe the linker with a link test before adding the flag so the build works with both GNU ld and lld. When the linker does not accept the option the symbol versioning provided by the explicit version script is still applied; only the implicit default version is dropped.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -135,7 +135,16 @@ AC_SUBST(JSON_BSYMBOLIC_LDFLAGS)
 
 # Enable symbol versioning on GNU libc
 JSON_SYMVER_LDFLAGS=
-AC_CHECK_DECL([__GLIBC__], [JSON_SYMVER_LDFLAGS=-Wl,--default-symver])
+AC_CHECK_DECL([__GLIBC__], [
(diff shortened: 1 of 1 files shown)


Zer0-One:
If your linker supports version scripts, why not just drop `--default-symver` altogether instead of jumping through hoops to check compatibility?

kraj:
> If your linker supports version scripts, why not just drop `--default-symver` altogether instead of jumping through hoops to check compatibility?

lld does support symbol versioning via explicit versioning scripts no fallbacks like --default-symver.

symbol versioning was added via ca80d5127e56dd2ee056fdf3f4204017b78b89e0 and I am not sure if jansson explicitly specified version script, in which case, removing this option would regress and cause glibc+gnu ld to not generate versioning symbols. for glibc+lld the option is to not use versioning.

